PR #4, opened 2024-11-18
Add middleware unit tests and natspec

NindoK:


girazoki:
@NindoK do we need the coverage to filter out contracts in script?

NindoK:
> @NindoK do we need the coverage to filter out contracts in script?

No I actually need to test those scripts. It will be done in the integration tests PR

nanocryk (approved):

PR #7, opened 2024-11-18
Add Demo script and Integration tests

NindoK:


src/middleware/Middleware.sol
```diff
@@ -402,24 +411,45 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
+            if (!_wasActiveAt(enabledTime, disabledTime, epochStartTs)) {
+                continue;
+            }
+
+            activeOperators[valIdx++] = operator;
         }
 
-        if (epochStartTs > Time.timestamp()) {
-            revert Middleware__InvalidEpoch();
+        assembly {
```
0xvgg, line 436: why through mstore

src/middleware/Middleware.sol
```diff
@@ -345,7 +352,6 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
         if (params.totalOperatorStake < amount) {
             revert Middleware__TooBigSlashAmount();
         }
-
         // simple pro-rata slasher
         for (uint256 i; i < s_vaults.length(); ++i) {
```
0xvgg, line 356: maybe change uint256 to smaller type as there is no way there will be 115792089237316195423570985008687907853269984665640564039457584007913129639935 vaults connected to tanssi. find something that maxes at 1024

src/middleware/Middleware.sol
```diff
@@ -402,24 +411,45 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
-    ) private view returns (uint256 totalStake) {
+    ) external view returns (address[] memory activeOperators) {
         uint48 epochStartTs = getEpochStartTs(epoch);
 
-        // for epoch older than i_slashingWindow total stake can be invalidated (use cache)
-        if (epochStartTs < Time.timestamp() - i_slashingWindow) {
-            revert Middleware__TooOldEpoch();
+        activeOperators = new address[](s_operators.length());
+        uint256 valIdx = 0;
+        for (uint256 i; i < s_operators.length(); ++i) {
```
0xvgg, line 425: same thing here - uint236 is too much - find one related to 1024

src/middleware/Middleware.sol
```diff
@@ -428,20 +458,40 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
+        for (uint256 j; j < s_vaults.length(); ++j) {
+            (address vault, uint48 vaultEnabledTime, uint48 vaultDisabledTime) = s_vaults.atWithTimes(j);
+
+            // just skip the vault if it was enabled after the target epoch or not enabled
+            if (!_wasActiveAt(vaultEnabledTime, vaultDisabledTime, epochStartTs)) {
+                continue;
+            }
+            uint256 operatorStake = 0;
+            for (uint96 k = 0; k < s_subnetworksCount; ++k) {
+                operatorStake += IBaseDelegator(IVault(vault).delegator()).stakeAt(
```
0xvgg, line 486: overflow?

src/middleware/Middleware.sol
NindoK: It's to shrink down the array and avoid wasting memory. It's how symbiotic does it too.

src/middleware/Middleware.sol
NindoK: If there's not a particular reason to decrease it, it's better to leave uint256 as it is costs less gas

src/middleware/Middleware.sol
NindoK: Same as above

src/middleware/Middleware.sol
NindoK: Don't think it's economically/math feasible. But it's how Symbiotic does it too

NindoK: Also we have a network max limit, so there is no possibility for an overflow if the network doesn't allow it

0xvgg (approved):

PR #11, opened 2024-12-16
Added benchmarking test for sorting array

NindoK:


nanocryk (approved):

nanocryk (changes requested):
Add unit test for the sorting function:
- Create an unsorted array
- Call sort function
- Compare neighbors to verify they are sorted

src/middleware/Middleware.sol
```diff
@@ -466,6 +466,41 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
         }
     }
 
+    function quickSort(ValidatorData[] memory arr, int256 left, int256 right) public pure {
+        int256 i = left;
```
ParthDesai, line 470: Is there a library for this?

ParthDesai (changes requested):
Please investigate whether we can use ready made library for quick sort.

PR #14, opened 2024-12-20
Add rewards

NindoK:
This PR:
- [ ] refactors the repo to divide between contracts and interfaces more clearly
- [ ] Introduces rewards for both operators and stakers.
- [ ] Rewards can be distributed only from the middleware, which then goes to operator via merkle root.
- [ ] Operator rewards control and send tokens on claim to be distributed to stakers
- [ ] Introduced multiple tests to reach 100% coverage for all contracts


src/contracts/middleware/Middleware.sol
```diff
@@ -38,73 +38,46 @@ import {Subnetwork} from "@symbiotic/contracts/libraries/Subnetwork.sol";
...
-        address operator;
-        uint256 totalOperatorStake;
-        uint256 slashAmount;
-    }
-
-    struct OperatorVaultPair {
-        address operator;
-        address[] vaults;
-    }
+    uint48 private constant INSTANT_SLASHER_TYPE = 0;
```
0xvgg, line 58: why not enum? it's 2 different values of the same type SlashType

src/contracts/middleware/Middleware.sol
```diff
@@ -310,14 +288,31 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
-    //     // validate signatures
-    //     // validate payload
-    //     // process payload
-    // }
+    function distributeRewards(
+        bytes calldata data
+    ) external onlyGateway {
+        (
+            uint256 timestamp,
+            , // Dunno what to do with it xD
```
0xvgg, line 306: leave "TODO:  xxxx" of any future work

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,143 @@
...
+        uint256 amount,
+        uint256 totalPointsToken,
+        bytes32 root
+    ) external nonReentrant onlyMiddleware {
+        if (amount > 0) {
+            uint256 balanceBefore = IERC20(i_token).balanceOf(address(this));
+            IERC20(i_token).safeTransferFrom(msg.sender, address(this), amount);
+            amount = IERC20(i_token).balanceOf(address(this)) - balanceBefore;
+
+            if (amount == 0) {
```
0xvgg, line 78: confusing. use another variable, don't override variables.

0xvgg: the block is if (amount > 0), but here you check for if (amount == 0) which shouldn't happen, but you have overrideen the input var - no,no

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,143 @@
...
+                revert ODefaultOperatorRewards__InsufficientTransfer();
+            }
+
+            s_balance[epoch].amount = amount;
+            s_balance[epoch].tokensPerPoint = amount / totalPointsToken; // To change the math. Check it's not zero
```
0xvgg, line 83: add TODO for any future work

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,143 @@
...
+        uint32 totalPointsClaimable,
+        bytes32[] calldata proof,
+        bytes calldata data
+    ) external nonReentrant returns (uint256 amount) {
+        bytes32 root_ = s_epochRoot[epoch];
+        if (root_ == bytes32(0)) {
+            revert ODefaultOperatorRewards__RootNotSet();
+        }
+
+        // This should be double checked
```
0xvgg, line 106: add TODO for any future work

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,143 @@
...
+        ) {
+            revert ODefaultOperatorRewards__InvalidProof();
+        }
+
+        address recipient = SimpleKeyRegistry32(
+            INetworkMiddlewareService(i_networkMiddlewareService).middleware(i_network)
+        ).getOperatorByKey(operatorKey);
+
+        uint256 claimed_ = s_claimed[epoch][recipient]; // this can beecome a bool.
+        amount = totalPointsClaimable * s_balance[epoch].tokensPerPoint * 10 ** 18; //Is it fine to use 10**18 here? Should we get the decimals for each token?
```
0xvgg, line 120: add TODO for any future work

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,143 @@
...
+        if (amount <= claimed_) {
+            revert ODefaultOperatorRewards__InsufficientTotalClaimable();
+        }
+
+        s_claimed[epoch][recipient] = amount;
+
+        //!Comment 1: Math here is important. Please double check if this is what we want!!
```
0xvgg, line 127: if it's important, explain the intent of the math. we want: XXXX

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,143 @@
...
+        uint256 operatorAmount = amount.mulDiv(20, 100); // 20% of the rewards to the operator
```
0xvgg, line 128: this should be configurable by the owner of the middleware. create variable and a setter guarded by onlyOwner

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,143 @@
...
+            )
+        ) {
+            revert ODefaultOperatorRewards__InvalidProof();
+        }
+
+        address recipient = SimpleKeyRegistry32(
+            INetworkMiddlewareService(i_networkMiddlewareService).middleware(i_network)
+        ).getOperatorByKey(operatorKey);
+
+        uint256 claimed_ = s_claimed[epoch][recipient]; // this can beecome a bool.
```
0xvgg, line 119: explain why it can become a bool

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -0,0 +1,405 @@
...
+        uint256 rewardsToClaim = Math.min(maxRewards, rewardsPerEpoch.length - rewardIndex);
+        uint48 epochTs = getEpochStartTs(epoch);
+        for (uint256 i; i < rewardsToClaim;) {
+            uint256 rewardAmount = rewardsPerEpoch[rewardIndex];
+
+            amount += IVault(s_vault).activeSharesOfAt(account, epochTs, new bytes(0)).mulDiv(
+                rewardAmount, _s_activeSharesCache[epoch]
+            );
+
+            unchecked {
```
0xvgg, line 239: why ++i is not in the for statement?

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -0,0 +1,405 @@
...
+            revert ODefaultStakerRewards__NoRewardsToClaim();
+        }
+
+        if (activeSharesOfHints.length == 0) {
+            activeSharesOfHints = new bytes[](rewardsToClaim);
+        } else if (activeSharesOfHints.length != rewardsToClaim) {
+            revert ODefaultStakerRewards__InvalidHintsLength();
+        }
+
+        //!Comment 1 Should we add that rewards are gonna expire after 1 month?
```
0xvgg, line 329: add TODO

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -0,0 +1,405 @@
...
+        for (uint256 i; i < rewardsToClaim;) {
+            uint256 rewardAmount = rewardsPerEpoch[rewardIndex];
+            amount += IVault(s_vault).activeSharesOfAt(msg.sender, epochTs, activeSharesOfHints[i]).mulDiv(
+                rewardAmount, _s_activeSharesCache[epoch]
+            );
+
+            unchecked {
+                ++i;
+                ++rewardIndex;
+                rewardsPerEpoch;
```
0xvgg, line 360: what does rewardsPerEpoch; do?

src/interfaces/rewarder/IODefaultStakerRewards.sol
```diff
@@ -0,0 +1,239 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
0xvgg, line 2: 0.8.25 - align solidity version for the whole project

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+    uint256 public constant AMOUNT_TO_DISTRIBUTE = 100 ether;
+    uint32 public constant AMOUNT_TO_CLAIM = 20;
+    uint32 public constant TOKEN_PER_POINT = 1;
+    uint256 public constant EXPECTED_CLAIMABLE = uint256(AMOUNT_TO_CLAIM) * uint256(TOKEN_PER_POINT) * 10 ** 18;
+    uint256 public constant ADMIN_FEE = 800; // 8%
+    bytes32 public constant REWARDS_ROOT = 0x4b0ddd8b9b8ec6aec84bcd2003c973254c41d976f6f29a163054eec4e7947810;
+    bytes32 public constant ALICE_KEY = 0x0404040404040404040404040404040404040404040404040404040404040404;
+    bytes32 public constant BOB_KEY = 0x0505050505050505050505050505050505050505050505050505050505050505;
+
+    bytes public constant REWARDS_ADDITIONAL_DATA =
```
0xvgg, line 74: explain what these values represent

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+    event DistributeRewards(uint48 indexed epoch, bytes32 indexed root);
+
+    uint48 public constant NETWORK_EPOCH_DURATION = 6 days;
+    uint48 public constant SLASHING_WINDOW = 7 days;
+    uint256 public constant AMOUNT_TO_DISTRIBUTE = 100 ether;
+    uint32 public constant AMOUNT_TO_CLAIM = 20;
+    uint32 public constant TOKEN_PER_POINT = 1;
+    uint256 public constant EXPECTED_CLAIMABLE = uint256(AMOUNT_TO_CLAIM) * uint256(TOKEN_PER_POINT) * 10 ** 18;
+    uint256 public constant ADMIN_FEE = 800; // 8%
+    bytes32 public constant REWARDS_ROOT = 0x4b0ddd8b9b8ec6aec84bcd2003c973254c41d976f6f29a163054eec4e7947810;
```
0xvgg, line 70: explain what this root represents

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+    bytes32 public constant ALICE_KEY = 0x0404040404040404040404040404040404040404040404040404040404040404;
```
0xvgg, line 71: can we have dynamic key provided by the testing framework?

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+        vm.startPrank(bob);
+        operatorRegistry.registerOperator();
+        operatorNetworkOptIn.optIn(tanssi);
+
+        vm.startPrank(tanssi);
+        middleware.registerOperator(alice, ALICE_KEY);
+        middleware.registerOperator(bob, BOB_KEY);
+
+        vm.startPrank(address(middleware));
+        feeToken = new MockFeeToken("Test", 100); //Extreme but to test that's a possibility
```
0xvgg, line 150: what is extreme? explain

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+    function _distributeRewards(uint48 epoch, uint256 amount) public {
+        vm.startPrank(address(middleware));
+        operatorRewards.distributeRewards(epoch, amount, amount, REWARDS_ROOT);
+        vm.stopPrank();
+    }
+
+    function _generateValidProof() internal pure returns (bytes32[] memory) {
+        bytes32[] memory proof = new bytes32[](1);
+        // Create a valid proof that matches the root we set
+        proof[0] = 0x27e610a11a547f210646001377ae223bc6bce387931f8153624d21f6478512d2;
```
0xvgg, line 211: what this proof means? is it somehow relates to REWARDS_ROOT? why not in const

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+        operatorRewards.claimRewards(ALICE_KEY, epoch, AMOUNT_TO_CLAIM, proof, REWARDS_ADDITIONAL_DATA);
+    }
+
+    function testClaimRewardsInvalidProof() public {
+        uint48 epoch = 0;
+        vm.warp(NETWORK_EPOCH_DURATION);
+        _distributeRewards(epoch, AMOUNT_TO_DISTRIBUTE);
+        bytes32[] memory proof = new bytes32[](1);
+        // Create a valid proof that matches the root we set
+        proof[0] = 0xffe610a11a547f210646001377ae223bc6bce387931f8153624d21f6478512d2;
```
0xvgg, line 318: what this proof means? is it somehow relates to REWARDS_ROOT? why not in const

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+    }
+
+    function testClaimRewardsWithTooHighAdminFee() public {
+        uint48 epoch = 0;
+        vm.warp(NETWORK_EPOCH_DURATION);
+        _mockVaultActiveSharesStakeAt(epoch, true, true);
+        _distributeRewards(epoch, AMOUNT_TO_DISTRIBUTE);
+
+        bytes memory badRewardsData =
+            hex"00000000000000000000000000000000000000000000000000000000000000050000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000";
```
0xvgg, line 353: why is it bad? what did you change?

0xvgg: if this is broken REWARDS_ADDITIONAL_DATA, then make it such, e.g. REWARDS_ADDITIONAL_DATA[12] = 4; which will override a particular byte of the data and it's clear that you're breaking it

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+            tanssi,
+            address(vaultFactory),
+            address(networkMiddlewareService),
+            epochStartTs,
+            NETWORK_EPOCH_DURATION,
+            address(feeToken),
+            params
+        );
+
+        bytes32 slot = keccak256(abi.encode(epoch, uint256(5))); // 5 is mapping slot number
```
0xvgg, line 697: 5 is mapping slot number of what?

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+        vm.store(address(stakerRewards), slot, bytes32(uint256(AMOUNT_TO_DISTRIBUTE / 10)));
```
0xvgg, line 698: why it's 10 here?

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+        stakerRewards.distributeRewards(epoch, AMOUNT_TO_CLAIM, REWARDS_ADDITIONAL_DATA);
+    }
+
+    //**************************************************************************************************
+    //                                      claimRewards
+    //**************************************************************************************************
+
+    function _setSRewardsMapping() private {
+        // For s_rewards[0] = [10 ether]
+        bytes32 slot = keccak256(abi.encode(uint48(0), uint256(2))); // 1 is mapping slot
```
0xvgg, line 752: why you're working with slots? can we have a view method?

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+        stakerRewards.claimRewards(alice, epoch, CLAIM_REWARDS_ADDITIONAL_DATA);
+    }
+
+    function testClaimStakerRewardsInvalidHintsLength() public {
+        uint48 epoch = 0;
+
+        _setSRewardsMapping();
+
+        bytes memory claimRewardsAdditionalDataWrong =
+            hex"0000000000000000000000000000000000000000000000056bc75e2d6310000000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000020b10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf60000000000000000000000000000000000000000000000000000000000000020b10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6";
```
0xvgg, line 848: if this is broken REWARDS_ADDITIONAL_DATA, then make it such, e.g. REWARDS_ADDITIONAL_DATA[12] = 4; which will override a particular byte of the data and it's clear that you're breaking it

test/unit/Rewards.t.sol
```diff
@@ -0,0 +1,964 @@
...
+            abi.encodeWithSelector(
+                IVaultStorage.activeSharesOfAt.selector,
+                alice,
+                epochTs,
+                hex"b10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6"
+            ),
+            abi.encode(AMOUNT_TO_DISTRIBUTE / 10)
+        );
+        bytes memory claimRewardsAdditionalDataFake =
+            hex"0000000000000000000000000000000000000000000000056bc75e2d631000000000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000020b10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6";
```
0xvgg, line 876: if this is broken REWARDS_ADDITIONAL_DATA, then make it such, e.g. REWARDS_ADDITIONAL_DATA[12] = 4; which will override a particular byte of the data and it's clear that you're breaking it

0xvgg: better yet, create a helper method _scrambledata(data)

test/unit/Rewards.t.sol
0xvgg: better yet, create a helper method _scrambledata(data)

test/unit/Rewards.t.sol
0xvgg: better yet, create a helper method _scrambledata(data)

0xvgg (changes requested):

test/unit/Rewards.t.sol
NindoK: Because I'm manipulating the state and mocking values. Otherwise I need to follow the whole flow from distributeRewards to claimRewards and for every scenario change these values and it would become very tedious . In thsi way I'm creating the exact scenario that I want

test/unit/Rewards.t.sol
NindoK: I'm saving 10 ethers inside the value. It's to say this person can claim only 1/10 of the amount distributed

test/unit/Rewards.t.sol
NindoK: It's the storage slot number of the variable `_s_activeSharesCache`

NindoK: Added comments to avoid confusion

src/contracts/rewarder/ODefaultOperatorRewards.sol
NindoK: Also I don't think changing would make sense. The code is highly optimized as follow and it makes sense. Reason being:

1. Amount is the amount that should be distributed
2. If the amount after the transfer is not the same amount should differ somehow
3. If it's 0 that's critical and should never happen
4. We should not check for different value, because there are tokens that have fees on transfer

src/contracts/rewarder/ODefaultStakerRewards.sol
NindoK: Because it's an optimization since we know it can't go past a specific value

src/contracts/rewarder/ODefaultStakerRewards.sol
NindoK: Left it there by mistake. Nice catch

src/interfaces/rewarder/IODefaultStakerRewards.sol
NindoK: It's a Library hence it's better to leave it as ^0.8.0. Then the contract is at 0.8.25

test/unit/Rewards.t.sol
NindoK: The fee. It's just to test an extreme scenario in tests in which the final amount is 0 and thus the revert is triggered

test/unit/Rewards.t.sol
NindoK: Done

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * @inheritdoc IODefaultOperatorRewards
+     */
+    mapping(uint48 epoch => BalancePerEpoch balancePerEpoch) public s_balance;
+
+    /**
+     * @inheritdoc IODefaultOperatorRewards
+     */
+    mapping(uint48 epoch => mapping(address account => uint256 amount)) public s_claimed;
```
girazoki, line 70: ah I see where you are going. You are considering this epoch to be the network epoch right? and for every "network epoch" you will keep adding when we receive rewards corresponding to that epoch, and then substract when people claim. Am I right here?

Please let me know

girazoki: But this is not going to work I believe. Here is the reason:

- we are sending the rewards with: `total_inflated`, `total_points`, `tanssi_epoch`, `middleware_network_epoch` and `reward_root`. Remember the last 2 can be different, one can be larger than the other for instance. What you know for sure is that the rewards will be sent once per `tanssi_epoch`.

- You dont accumulate rewards per operator when you receive that information. You just get the root. That  means that this variable needs to be a boolean and the epoch has to be the `tanssi_epoch` necessarily

src/contracts/middleware/Middleware.sol
```diff
@@ -310,14 +285,31 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
+        s_operatorRewards = IODefaultOperatorRewards(operatorRewardsAddress);
+
+        emit RewardsContractsSet(stakerRewardsAddress, operatorRewardsAddress);
+    }
+
+    function distributeRewards(
+        bytes calldata data
+    ) external onlyGateway {
+        (
+            uint256 timestamp,
```
girazoki, line 302: image

Yes you are missing the era index. This is the tanssi-epoch. Essentially, everything needs to be indexed by that

src/contracts/middleware/Middleware.sol
```diff
@@ -38,73 +38,43 @@ import {Subnetwork} from "@symbiotic/contracts/libraries/Subnetwork.sol";
...
-    uint48 public immutable i_slashingWindow;
 
-    uint48 public immutable i_startTime;
+    uint256 public s_subnetworksCount;
 
-    uint48 private constant INSTANT_SLASHER_TYPE = 0;
-    uint48 private constant VETO_SLASHER_TYPE = 1;
+    mapping(uint48 epoch => uint256 amount) public s_totalStakeCache;
+    mapping(uint48 epoch => bool) public s_totalStakeCached;
+    mapping(uint48 epoch => mapping(address operator => uint256 amount)) public s_operatorStakeCache;
```
girazoki, line 71: This needs to be the tanssi era index. Everything will be indexed by that. That is the only "attachment" you have to tanssi whatsoever. That means that we also need a mapping between tanssi-era-index -> network_epoch

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,183 @@
...
+            revert ODefaultOperatorRewards__InvalidProof();
+        }
+
+        address recipient = SimpleKeyRegistry32(
+            INetworkMiddlewareService(i_networkMiddlewareService).middleware(i_network)
+        ).getOperatorByKey(operatorKey);
+
+        uint256 claimed_ = s_claimed[epoch][recipient]; //TODO: This can become a bool. uint256 it's better even UI wise as we would know exactly how much was claimed instead of querying 2 times the contract.
+
+        amount = totalPointsClaimable * s_balance[epoch].tokensPerPoint * 10 ** 18; //! TODO Is it fine to use 10**18 here? We are saying that are all IERC20 with 18 decimals. What if we use USDC as collateral? Should we get the decimals for each token?
```
girazoki, line 146: the balance will already be with the appropriate decimals, why are we multiplying by 10^18?

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,183 @@
...
+            )
+        ) {
+            revert ODefaultOperatorRewards__InvalidProof();
+        }
+
+        address recipient = SimpleKeyRegistry32(
+            INetworkMiddlewareService(i_networkMiddlewareService).middleware(i_network)
+        ).getOperatorByKey(operatorKey);
+
+        uint256 claimed_ = s_claimed[epoch][recipient]; //TODO: This can become a bool. uint256 it's better even UI wise as we would know exactly how much was claimed instead of querying 2 times the contract.
```
girazoki, line 144: Not having a boolean allows someont to claim twice if I am not mistaken

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -0,0 +1,399 @@
...
+            revert ODefaultStakerRewards__NotVault();
+        }
+        s_vault = vault;
+    }
+
+    /**
+     * @inheritdoc IODefaultStakerRewards
+     */
+    function claimable(
+        uint48 epoch,
```
girazoki, line 218: again, tanssi-era

src/contracts/middleware/Middleware.sol
```diff
@@ -433,9 +424,9 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
             return;
         }
         uint256 slasherType = IEntity(slasher).TYPE();
-        if (slasherType == INSTANT_SLASHER_TYPE) {
+        if (slasherType == uint256(SlasherType.INSTANT)) {
```
0xvgg, line 427: can we avoid this casting by making SlasherType.INSTANT uint256?

src/contracts/rewarder/ODefaultOperatorRewards.sol
0xvgg: it's not a matter of maount being amount, it's a matter of overriding the input param - which is not nice and confusing.

src/contracts/rewarder/ODefaultOperatorRewards.sol
0xvgg: let's not consider UI right now. UI can always do ETL to transform the values to the needed format

src/contracts/rewarder/ODefaultOperatorRewards.sol
0xvgg: agree, 
s_balance[epoch].amount was set with amount, which we received in distributeRewards
function distributeRewards(
        uint48 epoch,
        uint256 amount,

but @NindoK overrode the value of amount, which is a separate comment. in anyway, s_balance should be in the correct type set in distributeRewards

0xvgg (changes requested):

src/contracts/middleware/Middleware.sol
NindoK: Nope

src/contracts/middleware/Middleware.sol
NindoK: I think these shouldn't be touched as they are relative to ethereum epoch. It's not related to rewards or anything

src/contracts/rewarder/ODefaultStakerRewards.sol
NindoK: I'm thinking this should remain epoch since for stakers they have the eraindex intrinsically in the array of rewards. Let's talk tomorrow about it real quick

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,183 @@
...
+
+        amount = totalPointsClaimable * s_balance[epoch].tokensPerPoint * 10 ** 18; //! TODO Is it fine to use 10**18 here? We are saying that are all IERC20 with 18 decimals. What if we use USDC as collateral? Should we get the decimals for each token?
+
+        if (amount <= claimed_) {
+            revert ODefaultOperatorRewards__InsufficientTotalClaimable();
+        }
+
+        s_claimed[epoch][recipient] = amount;
```
ParthDesai, line 152: shouldn't you add this amount to previous value of `s_claimed`?

src/contracts/middleware/Middleware.sol
```diff
@@ -388,7 +395,7 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
                     || params.epochStartTs >= Time.timestamp()
             ) {
                 emit InvalidSlashTimeframe(epoch, operator, amount);
```
0xvgg, line 397: InvalidSlashTimeframe error is only for specific vault among s_vaults, so it will be unclear which one. add to InvalidSlashTimeframe vault address/id

src/contracts/middleware/Middleware.sol
```diff
@@ -475,15 +480,20 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
     }
 
     /**
-     * @notice Gets the operators' keys for latest epoch
-     * @return keys Array of operator keys
+     * @inheritdoc IMiddleware
      */
+    // TODO: THIS MUST BE RESTRICTED EITHER TO OWNER OR SOMEONE ELSE
     function sendCurrentOperatorsKeys() external returns (bytes32[] memory keys) {
+        if (address(s_gateway) == address(0)) {
```
0xvgg, line 487: yes, this is ownerOnly

why do we have operatorKeys setter as array? each operator should set his own key, no?

0xvgg (changes requested):

src/contracts/middleware/Middleware.sol
```diff
@@ -433,9 +440,9 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
         }
         uint256 slasherType = IEntity(slasher).TYPE();
-        if (slasherType == INSTANT_SLASHER_TYPE) {
+        if (slasherType == uint256(SlasherType.INSTANT)) {
             ISlasher(slasher).slash(subnetwork, operator, amount, timestamp, new bytes(0));
-        } else if (slasherType == VETO_SLASHER_TYPE) {
+        } else if (slasherType == uint256(SlasherType.VETO)) {
             IVetoSlasher(slasher).requestSlash(subnetwork, operator, amount, timestamp, new bytes(0));
         } else {
             revert Middleware__UnknownSlasherType();
```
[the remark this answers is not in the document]
girazoki: same here, changing this in my PR. so dont worry about this

src/contracts/middleware/Middleware.sol
girazoki: why this has to be restricted? I think it can be sent by anyone no?

NindoK: Where is this setter?

src/contracts/middleware/Middleware.sol
```diff
@@ -475,15 +480,20 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
     }
 
     /**
-     * @notice Gets the operators' keys for latest epoch
-     * @return keys Array of operator keys
+     * @inheritdoc IMiddleware
      */
+    // TODO: THIS MUST BE RESTRICTED EITHER TO OWNER OR SOMEONE ELSE
```
girazoki, line 485: remove TODO

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,235 @@
...
+    function distributeRewards(
+        uint48 epoch,
+        uint48 eraIndex,
+        uint256 amount,
+        uint256 totalPointsToken,
+        bytes32 root
+    ) external nonReentrant onlyMiddleware {
+        if (amount > 0) {
+            uint256 balanceBefore = IERC20(i_token).balanceOf(address(this));
+            IERC20(i_token).safeTransferFrom(msg.sender, address(this), amount);
```
girazoki, line 111: what is this? this is wrong no?

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -0,0 +1,391 @@
...
+        bytes calldata data
+    ) external override nonReentrant onlyRole(OPERATOR_REWARDS_ROLE) {
+        // maxAdminFee - the maximum admin fee to allow
+        // activeSharesHint - a hint index to optimize `activeSharesAt()` processing
+        // activeStakeHint - a hint index to optimize `activeStakeAt()` processing
+        (uint256 maxAdminFee, bytes memory activeSharesHint, bytes memory activeStakeHint) =
+            abi.decode(data, (uint256, bytes, bytes));
+
+        uint48 epochTs = getEpochStartTs(epoch);
+        if (epochTs >= Time.timestamp()) {
```
girazoki, line 247: comment

girazoki (approved):
Please let's add comments on the main logics of `claimRewards` and `distributeRewards`, in both rewarders

src/contracts/rewarder/ODefaultOperatorRewards.sol
girazoki: it's correct

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -0,0 +1,235 @@
...
+
+        // (1-s_operatorShare)% of the rewards to the stakers
+        uint256 stakerAmount = amount - operatorAmount;
+
+        // On every claim send s_operatorShare% of the rewards to the operator
+        // And then distribute rewards to the stakers
+        // This is gonna send (1-s_operatorShare)% of the rewards
+        IERC20(i_token).safeTransfer(recipient, operatorAmount);
+
+        _distributeRewardsToStakers(input.epoch, input.eraIndex, stakerAmount, recipient, middlewareAddress, input.data);
```
girazoki, line 179: epoch should not be in the input, but rather, we should fetch it from storage. Otherwise it can be gamed

PR #15, opened 2025-01-07
Send operators data through primary channel

nanocryk:
Refactor code to allow sending message to a given channel ID. Now operators are sent to channel PRIMARY_GOVERNANCE_CHANNEL_ID.


src/snowbridge-override/Gateway.sol
```diff
@@ -576,10 +579,6 @@ contract Gateway is IOGateway, IInitializable, IUpgradable {
         Ticket memory ticket
     ) internal {
         ChannelID channelID = ticket.dest.into();
-        Channel storage channel = _ensureChannel(channelID);
-
-        // Ensure outbound messaging is allowed
-        _ensureOutboundMessagingEnabled(channel);
 
         // // Destination fee always in DOT
         // uint256 fee = _calculateFee(ticket.costs);
```
NindoK, line 584: Can you revert back then the _submitOutbound function to its prior version?

src/snowbridge-override/Gateway.sol
```diff
@@ -518,9 +518,12 @@ contract Gateway is IOGateway, IInitializable, IUpgradable {
...
     }
 
-    function sendOperatorsData(bytes32[] calldata data, ParaID destinationChain) external {
-        Ticket memory ticket = Operators.encodeOperatorsData(data, destinationChain);
-        _submitOutbound(ticket);
+    function sendOperatorsData(
+        bytes32[] calldata data
+    ) external {
+        ParaID dummyChain = ParaID.wrap(0);
+        Ticket memory ticket = Operators.encodeOperatorsData(data, dummyChain);
```
NindoK, line 525: At this point take out the dummyChain and take out dest in the ticket.

src/snowbridge-override/Operators.sol
```diff
@@ -41,8 +40,9 @@ library Operators {
             revert Operators__OperatorsLengthTooLong();
         }
 
-        ticket.dest = dest;
-        //TODO For now mock it to 0
+        // TODO: This is a type from Snowbridge, do we want our own simplified Ticket type?
```
NindoK, line 43: Does it make sense? For now this is our limited use case, but probably later on if we need it would make sense def

girazoki (approved):

PR #23, opened 2025-01-16
Add slashes as a percentage and change execution to be non-failable with try catch

girazoki:
This PR adds the slashes as a percentage (parts per billion, as will be transferred from tanssi) and additionally makes the slashing non-failable with try-catch

Also changes slashes to receive the 32 bytes operator key instead


src/middleware/Middleware.sol
```diff
@@ -362,20 +365,16 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
+    function slash(uint48 epoch, address operator, uint256 percentage) public onlyOwner updateStakeCache(epoch) {
         SlashParams memory params;
         params.epochStartTs = getEpochStartTs(epoch);
         params.operator = operator;
-        params.slashAmount = amount;
+        params.slashPercentage = percentage;
 
         params.totalOperatorStake = getOperatorStake(operator, epoch);
-
-        if (params.totalOperatorStake < amount) {
```
NindoK, line 376: I don't like this taken out. What you think if we emit an event and return instead?

src/middleware/Middleware.sol
```diff
@@ -362,20 +365,16 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
+     * @param percentage Percentage to slash, represented as parts per billion.
      */
     //INFO: this function can be made external. To check if it is possible to make it external
-    function slash(uint48 epoch, address operator, uint256 amount) public onlyOwner updateStakeCache(epoch) {
+    function slash(uint48 epoch, address operator, uint256 percentage) public onlyOwner updateStakeCache(epoch) {
         SlashParams memory params;
         params.epochStartTs = getEpochStartTs(epoch);
         params.operator = operator;
-        params.slashAmount = amount;
+        params.slashPercentage = percentage;
```
NindoK, line 375: We need a sanity check that this can't be higher than 100% or whatever maximum value we want to put

girazoki: yep let me do it

src/middleware/Middleware.sol
```diff
@@ -408,7 +399,9 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
             uint256 vaultStake = IBaseDelegator(IVault(vault).delegator()).stakeAt(
                 subnetwork, params.operator, params.epochStartTs, new bytes(0)
             );
-            uint256 slashAmount = (params.slashAmount * vaultStake) / params.totalOperatorStake;
+            // Slash percentage is already in parts per billion
+            // so we need to divide by a billion
+            uint256 slashAmount = params.slashPercentage.mulDiv(vaultStake, 1_000_000_000);
```
0xvgg, line 404: make 1_000_000_000 as const

src/middleware/Middleware.sol
```diff
@@ -430,16 +423,30 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
-            IVetoSlasher(slasher).requestSlash(subnetwork, operator, amount, timestamp, new bytes(0));
+            try IVetoSlasher(slasher).requestSlash(subnetwork, operator, amount, timestamp, new bytes(0)) {
+                return;
+            } catch Error(string memory _err) {
+                emit SlashFailure(_err, subnetwork, operator, amount, timestamp);
+            } catch (bytes memory _err) {
+                emit SlashFailure(_err, subnetwork, operator, amount, timestamp);
+            }
         } else {
-            revert Middleware__UnknownSlasherType();
```
0xvgg, line 442: this shuoldn't happen anyway, but at least emit event for monitoring

test/fork/Middleware.t.sol
```diff
@@ -428,9 +428,11 @@ contract MiddlewareTest is Test {
         assertEq(validators[2].stake, totalOperator3Stake + remainingOperator3Stake);
         //We calculate the amount slashable for only the operator2 since it's the only one that should be slashed. As a side effect operator3 will be slashed too since it's taking part in a NetworkRestake delegator based vault
         uint256 slashAmountSlashable = (SLASH_AMOUNT * remainingOperator2Stake) / totalOperator2Stake;
+        uint256 amountToSlash = 30 ether;
+        uint256 slashingFraction = amountToSlash.mulDiv(1_000_000_000, totalOperator2Stake);
```
0xvgg, line 432: 1_000_000_000 as const or local variable in test

test/integration/Middleware.t.sol
```diff
@@ -505,10 +509,24 @@ contract MiddlewareTest is Test {
         // We go directly to epochStart as it 100% ensure that the epoch is started and thus the slashing is invalid
         vm.warp(epochStartTs);
 
+        //InvalidCaptureTimestamp
+        bytes memory data = "\x26\xaf\x59\x22";
```
0xvgg, line 513: what is this? what does it mean?

test/integration/Middleware.t.sol
```diff
@@ -626,6 +658,7 @@ contract MiddlewareTest is Test {
 
         (uint256 totalOperator3StakeAfter,) =
             _calculateTotalOperatorStake(OPERATOR_STAKE * 2 * 2, activeStakeInVetoed, slashAmountSlashable3);
```
0xvgg, line 660: why to multiply by 4

test/unit/Middleware.t.sol
```diff
@@ -957,8 +963,10 @@ contract MiddlewareTest is Test {
         vm.warp(START_TIME + SLASHING_WINDOW + 1);
         uint48 currentEpoch = middleware.getCurrentEpoch();
 
+        // 50% of slashing
+        uint256 slashPercentage = 500_000_000;
         uint256 slashAmount = OPERATOR_STAKE / 2;
-        middleware.slash(currentEpoch, operator, slashAmount);
+        middleware.slash(currentEpoch, operator, slashPercentage);
 
         vm.warp(SLASHING_WINDOW * 2 + 1);
```
0xvgg, line 971: explain what *2 + 1 means

0xvgg (changes requested):

test/integration/Middleware.t.sol
girazoki: it's the seelctor of the event

test/integration/Middleware.t.sol
girazoki: it's bennys setup, we has more stake for one of the operators (operator 3)

src/middleware/Middleware.sol
girazoki: fixed

test/unit/Middleware.t.sol
NindoK: You are basically going after 2 slashing_window (i.e. slashing window is 7 days => 14 days) + 1 (ms).

test/unit/Middleware.t.sol
```diff
@@ -1012,9 +1020,14 @@ contract MiddlewareTest is Test {
         uint48 currentEpoch = middleware.getCurrentEpoch();
 
         uint256 slashAmount = OPERATOR_STAKE / 2;
-        vm.expectRevert(Middleware.Middleware__UnknownSlasherType.selector);
```
NindoK, line 1015: Take care because the test is for wrongType. If you have changed now the logic and this doesn't exist anymore, please change the test name.

test/unit/Middleware.t.sol
NindoK: Probably it's my mistake here. Should have been a bit more coherent with actual timing. This should mean in an actual epoch, hence should suffice `NETWORK_EPOCH_DURATION + SLASHING_WINDOW + 1`.

src/middleware/Middleware.sol
```diff
@@ -48,8 +49,10 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
     using EnumerableMap for EnumerableMap.AddressToUintMap;
     using MapWithTimeData for EnumerableMap.AddressToUintMap;
     using Subnetwork for address;
+    using Math for uint256;
 
-    event InvalidSlashTimeframe(uint48 indexed epoch, address indexed operator, uint256 indexed amount);
+    event SlashFailure(string stringFailure, bytes32 subnetwork, address operator, uint256 amount, uint48 timestamp);
```
NindoK, line 53: I'd put as indexed operator, amount and timestamp

src/middleware/Middleware.sol
```diff
@@ -48,8 +49,10 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
+    event SlashFailure(bytes bytesFailure, bytes32 subnetwork, address operator, uint256 amount, uint48 timestamp);
```
NindoK, line 54: Same as above

src/middleware/Middleware.sol
NindoK: I don't see the new event?

test/unit/Middleware.t.sol
girazoki: I did not. Check a bit below. I am asserting the event is thrown

src/middleware/Middleware.sol
girazoki: without subnetwork? ok

NindoK: You just have to put the indexed keyword on those that I said. This means that these values are "tracked" and we can easily query and filter by these values

girazoki: done

src/middleware/Middleware.sol
girazoki: done

src/middleware/Middleware.sol
```diff
@@ -429,16 +429,31 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
+                return;
+            } catch Error(string memory _err) {
+                emit SlashFailure(_err, subnetwork, operator, amount, timestamp);
+            } catch (bytes memory _err) {
+                emit SlashFailure(_err, subnetwork, operator, amount, timestamp);
+            }
         } else {
-            revert Middleware__UnknownSlasherType();
+            // There is no slasher type we can apply and we cannot revert
+            emit UnknownSlasherType();
```
0xvgg, line 455: add slasherType to the UnknownSlasherType event so that it's clear which type is unknown

0xvgg (changes requested):

test/unit/Middleware.t.sol
NindoK: Am I seeing at an older code? I don't see any error being thrown

Screenshot 2025-01-20 at 21 56 22

girazoki: because we dont throw an error now, but an event. 

image

src/middleware/Middleware.sol
girazoki: fixed

girazoki:
DONT merge. the function accepts a 160 byte address for slashing, but that is not the information we get

test/unit/Middleware.t.sol
girazoki: resolved as now we have the revert back

src/middleware/Middleware.sol
```diff
@@ -360,21 +363,29 @@ contract Middleware is SimpleKeyRegistry32, Ownable {
...
         SlashParams memory params;
-        params.epochStartTs = getEpochStartTs(epoch);
+        params.epochStartTs = epochStartTs;
         params.operator = operator;
-        params.slashAmount = amount;
+        params.slashPercentage = percentage;
 
         params.totalOperatorStake = getOperatorStake(operator, epoch);
-
-        if (params.totalOperatorStake < amount) {
```
NindoK, line 375: I think this if should remain. We don't want to slash for more than the actual operator stake. We need to calculate it later on though

test/integration/Middleware.t.sol
```diff
@@ -504,10 +510,44 @@ contract MiddlewareTest is Test {
         // We go directly to epochStart as it 100% ensure that the epoch is started and thus the slashing is invalid
         vm.warp(epochStartTs);
 
+        //InvalidCaptureTimestamp
+        bytes memory data = "\x26\xaf\x59\x22";
```
NindoK, line 514: Can you just put it as hex"26af5922" ?

test/integration/Middleware.t.sol
```diff
@@ -695,6 +759,55 @@ contract MiddlewareTest is Test {
...
+            _calculateTotalOperatorStake(OPERATOR_STAKE * 2, activeStakeInVetoed, 0);
+
+        assertEq(validators[1].stake, totalOperator2Stake);
+        //We need to assert like this instead of putting OPERATOR_STAKE * 2 * 2 because of the precision loss. We know that remainingOperator3Stake will be the same even for the other vault so we can just sum it.
+        assertEq(validators[2].stake, totalOperator3Stake + remainingOperator3Stake);
+
+        //We calculate the amount slashable for only the operator2 since it's the only one that should be slashed. As a side effect operator3 will be slashed too since it's taking part in a NetworkRestake delegator based vault
+        uint256 slashAmountSlashable = (SLASH_AMOUNT * remainingOperator2Stake) / totalOperator2Stake;
+
+        vm.prank(owner);
```
NindoK, line 783: When you have to do multiple call with the same address just use: `vm.startPrank` which is cleaner and explains that this specific area pertains to the owner logic/action

test/integration/Middleware.t.sol
```diff
@@ -695,6 +759,55 @@ contract MiddlewareTest is Test {
...
+
+        uint256 slashedAmount = 30 ether;
+        // We want to slash 30 ether, so we need to calculate what percentage
+        uint256 slashingFraction = slashedAmount.mulDiv(PARTS_PER_BILLION, totalOperator2Stake);
+
+        // Before slashing, we will change the operator2 key to something else, and prove we can still slash
```
NindoK, line 789: Add to the comment: this works since the operatorKey it's taken based on the timestamp. Hence when we update it here it's still not reflected in the current epoch.

NindoK (approved):

test/integration/Middleware.t.sol
girazoki: yes

girazoki: although this should not be relevant aynmore

src/middleware/Middleware.sol
girazoki: we are slashing based on percentage. so it's impossible that we slash more than totalOperatorStake if the percentage is less than 100% no?  and we have a check for that

NindoK: Oh yes you are right. Thanks

NindoK (approved):

PR #27, opened 2025-01-27
Added ability to receive a token address and based on that distribute rewards and claim rewards

NindoK:
This PR add:
- Possibility to reward based on different tokens
- Flow from middleware support token address to be sent with rewards
- Tests to make sure the flow is correct

### DIFF
--- a/src/contracts/rewarder/ODefaultStakerRewards.sol
+++ b/src/contracts/rewarder/ODefaultStakerRewards.sol
@@ -382,7 +382,6 @@ contract ODefaultStakerRewards is
         uint48 epoch,
         address tokenAddress
     ) external nonReentrant onlyRole(ADMIN_FEE_CLAIM_ROLE) {
-        // This implies that we have only one token address which should be the case initially.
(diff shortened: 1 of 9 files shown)


src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -112,17 +107,18 @@ contract ODefaultStakerRewards is
     /**
      * @inheritdoc IODefaultStakerRewards
      */
-    mapping(uint48 epoch => uint256[] rewards_) public s_rewards;
+    mapping(uint48 epoch => mapping(address tokenAddress => uint256[] rewards_)) public s_rewards;
```
girazoki, line 110: is this easy to iterate later on from a UI perspective?

NindoK: Yes as we will know which token we have from the gateway, hence they will keep a list of tokens we support or based on the vault you get the token and from there you get everything

girazoki (approved):
LGTM

nanocryk (approved):

PR #40, opened 2025-02-23
Fix ordering of variables in event. Took out unused events/errors

NindoK:
Minor PR to fix variables ordering error in DistributeRewards and take out unused code

### DIFF
--- a/test/integration/Rewards.t.sol
+++ b/test/integration/Rewards.t.sol
@@ -432,7 +432,7 @@ contract MiddlewareTest is Test {
     function _makeReportRewardsCommand(
         uint256 amount
     ) public returns (Command, bytes memory, address) {
-        uint256 timestamp = ONE_DAY * 3;
+        uint256 epoch = 1;
         uint256 eraIndex = 1;
         uint256 totalPointsToken = amount;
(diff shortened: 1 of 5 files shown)


test/integration/Rewards.t.sol
```diff
@@ -502,7 +502,7 @@ contract MiddlewareTest is Test {
 
         uint48 epoch = 1;
         vm.expectEmit(false, true, true, true);
-        emit IODefaultOperatorRewards.DistributeRewards(epoch, 0, tokenAddress, 1, amount, bytes32(uint256(1)));
+        emit IODefaultOperatorRewards.DistributeRewards(0, epoch, tokenAddress, 1, amount, bytes32(uint256(1)));
```
nanocryk, line 505: By looking at the interface order is epoch followed by eraIndex, while here you swapping them (unless variable "epoch" should be named "eraIndex" in this test?)

nanocryk (changes requested):

test/integration/Rewards.t.sol
NindoK: You are right 😂 I thought everything was wrong ahaha

nanocryk (approved):

PR #42, opened 2025-02-25
Initial non working Middleware

NindoK:
This PR aims to add the new middleware v2


script/DeployTanssiEcosystem.s.sol
```diff
@@ -308,6 +314,30 @@ contract DeployTanssiEcosystem is Script {
...
+        address _vaultRegistry,
+        address _operatorNetworkOptInService,
+        address _owner,
+        uint48 _epochDuration,
+        uint48 _slashingWindow
+    ) private returns (Middleware _middleware) {
+        Middleware _middlewareImpl = new Middleware();
+        _middleware = Middleware(address(new ERC1967Proxy(address(_middlewareImpl), "")));
+        address readHelper = address(new BaseMiddlewareReader());
+        Middleware(address(_middleware)).initialize(
```
steven2308, line 329: I think this recasting isn't needed.
```suggestion
        _middleware.initialize(
```

src/contracts/middleware/Middleware.sol
```diff
@@ -389,51 +235,46 @@ contract Middleware is SimpleKeyRegistry32, Ownable, IMiddleware {
...
+    //TODO use prebuilt function from SDK
     function slash(
         uint48 epoch,
         bytes32 operatorKey,
         uint256 percentage
     ) external onlyGateway updateStakeCache(epoch) {
-        uint48 epochStartTs = getEpochStartTs(epoch);
-        address operator = getOperatorByKey(operatorKey);
+        uint48 epochStartTs = getEpochStart(epoch);
+        address operator = operatorByKey(abi.encode(operatorKey));
```
steven2308, line 277: Feel `getOperatorByKey` was better name, also consistent with  the other getter

src/contracts/middleware/Middleware.sol
```diff
@@ -516,30 +351,14 @@ contract Middleware is SimpleKeyRegistry32, Ownable, IMiddleware {
     //                                      VIEW FUNCTIONS
     // **************************************************************************************************
 
-    /**
-     * @inheritdoc IMiddleware
-     */
+    // /**
+    //  * @inheritdoc IMiddleware
+    //  */
```
steven2308, line 356: I've seen many of these. I'm guessing either `IMiddleware` changed or we stopped inheriting from it. Shall we update/create a new interface?

src/contracts/middleware/Middleware.sol
```diff
@@ -389,51 +235,46 @@ contract Middleware is SimpleKeyRegistry32, Ownable, IMiddleware {
...
+        address[] memory _operators = _activeOperatorsAt(epochStartTs);
 
-        for (uint256 i; i < s_operators.length(); ++i) {
-            (address operator, uint48 enabledTime, uint48 disabledTime) = s_operators.atWithTimes(i);
-
-            // just skip operator if it was added after the target epoch or paused
-            if (!_wasActiveAt(enabledTime, disabledTime, epochStartTs)) {
-                continue;
-            }
+        for (uint256 i; i < _operators.length; ++i) {
```
steven2308, line 255: Not sure if this changed recently, but AFAIK it is more gas efficient to get the length before so it's read only once. There are multiple cases of it. I have the same doubt for `++i` not being in an unchecked block instead.

src/contracts/middleware/Middleware.sol
```diff
@@ -591,73 +405,56 @@ contract Middleware is SimpleKeyRegistry32, Ownable, IMiddleware {
...
-    /**
-     * @inheritdoc IMiddleware
-     */
+    // /**
+    //  * @inheritdoc IMiddleware
+    //  */
     function getOperatorVaults(
         address operator,
         uint48 epochStartTs
     ) public view returns (uint256 vaultIdx, address[] memory _vaults) {
```
steven2308, line 414: Any reason to use `_` on `_vaults`?

test/integration/Middleware.t.sol
```diff
@@ -969,7 +979,7 @@ contract MiddlewareTest is Test {
                     tanssi.subnetwork(0), _operator, depositAmount
                 );
             }
-            middleware.registerOperator(_operator, bytes32(uint256(i + 4)));
+            middleware.registerOperator(_operator, abi.encode(bytes32(uint256(i + 4))), address(0));
```
steven2308, line 982: Could we put this into a variable with a descriptive name. Pretty hard to know what this is about at first glance:
`abi.encode(bytes32(uint256(i + 4)))`

test/unit/Middleware.t.sol
```diff
@@ -119,16 +127,21 @@ contract MiddlewareTest is Test {
...
-            address(registry),
-            address(operatorNetworkOptInServiceMock),
-            owner,
-            NETWORK_EPOCH_DURATION,
-            SLASHING_WINDOW
+
+        Middleware middlewareImpl = new Middleware();
+        middleware = Middleware(address(new ERC1967Proxy(address(middlewareImpl), "")));
+        address readHelper = address(new BaseMiddlewareReader());
+        Middleware(address(middleware)).initialize(
```
steven2308, line 134:
```suggestion
        middleware.initialize(
```
I think casting is not needed here.

test/unit/Middleware.t.sol
```diff
@@ -119,16 +127,21 @@ contract MiddlewareTest is Test {
...
+            address(registry), // operatorRegistry
+            address(registry), // vaultRegistry
+            address(operatorNetworkOptInServiceMock), // operatorNetOptin
+            owner, // owner
+            NETWORK_EPOCH_DURATION, // epoch duration
+            SLASHING_WINDOW, // slashing window
+            readHelper // reader
         );
-        middleware.setGateway(address(gateway));
+        Middleware(address(middleware)).setGateway(address(gateway));
```
steven2308, line 144:
```suggestion
        middleware.setGateway(address(gateway));
```
Same here.

steven2308 (commented):
Minor stuff.

src/contracts/middleware/Middleware.sol
NindoK: I concur, but it comes from their `KeyManager256.sol` contract

src/contracts/middleware/Middleware.sol
NindoK: Yup let's add it in another PR since we need to change it in lots of places

steven2308 (approved):

PR #43, opened 2025-03-04
MD-927 Auto deploy staker rewards when registering shared vaults.

steven2308:
When registering a shared vault, a staker rewards contract is deployed automatically using the factory and the configured default parameters.


src/contracts/middleware/Middleware.sol
```diff
@@ -171,10 +177,35 @@ contract Middleware is
...
+            revert Middleware__VaultCannotHaveADefault();
+        }
+        s_stakerRewardsInitParams = params;
+    }
+
     // /**
     //  * @inheritdoc IMiddleware
     //  */
     //TODO this will be removed and done automatically when registering a vault first time by creating staker contract from factory and then setting the mapping in operators for vault <=> staker contract
+    // TODO: If we apply the suggested to do there will be now way to later change the staker rewards contract for a vault. Is that a problem?
```
steven2308, line 208: @NindoK WDYT? 
If not a problem I can do the change on this same PR

src/contracts/middleware/Middleware.sol
```diff
@@ -171,10 +177,35 @@ contract Middleware is
...
+        emit StakerRewardsFactorySet(stakerRewardsFactory);
+    }
+
+    function setStakerRewardsInitParams(
+        IODefaultStakerRewards.InitParams memory params
+    ) external checkAccess {
+        if (params.vault != address(0)) {
+            revert Middleware__VaultCannotHaveADefault();
+        }
+        s_stakerRewardsInitParams = params;
```
steven2308, line 201: Shall we emit and event here also?

src/contracts/middleware/Middleware.sol
```diff
@@ -539,4 +570,23 @@ contract Middleware is
...
     }
+
+    /**
+     * @inheritdoc SharedVaults
+     */
+    function _beforeRegisterSharedVault(
+        address sharedVault
+    ) internal virtual override {
+        // TODO: Shall we revert if no operator rewards are set? Looks like they should be before registering a shared vault.
+        // TODO: Shall we revert if no staker rewards factory is set? Makes sense to do if we remove the setStakerRewardsFactory function.
```
steven2308, line 581: @NindoK I think we should revert in both cases, specially if we remove the setter.

src/contracts/middleware/Middleware.sol
NindoK: That shouldn't be an issue since the staker rewards contract will be upgradeable

src/contracts/middleware/Middleware.sol
NindoK: For the former yes, use the modifier. For the latter no since it's gonna be immutable

script/DeployTanssiEcosystem.s.sol
```diff
@@ -315,27 +325,14 @@ contract DeployTanssiEcosystem is Script {
...
-        address _operatorNetworkOptInService,
-        address _owner,
-        uint48 _epochDuration,
-        uint48 _slashingWindow
+        IMiddleware.InitParams memory params
     ) private returns (Middleware _middleware) {
         Middleware _middlewareImpl = new Middleware();
         _middleware = Middleware(address(new ERC1967Proxy(address(_middlewareImpl), "")));
+
         address readHelper = address(new BaseMiddlewareReader());
```
NindoK, line 333: Check if readHelper != address(0) => then take that otherwise deploy it

script/DeployTanssiEcosystem.s.sol
```diff
@@ -274,31 +276,39 @@ contract DeployTanssiEcosystem is Script {
...
-            operatorRegistryAddress,
-            vaultRegistryAddress,
-            operatorNetworkOptInServiceAddress,
-            tanssi,
-            NETWORK_EPOCH_DURATION,
-            SLASHING_WINDOW
+
+        deployRewards = new DeployRewards();
+        (address stakerRewardsFactoryAddress,) = deployRewards.deployStakerRewardsFactoryContract(
+            vaultRegistryAddress, networkMiddlewareServiceAddress, 1 days, NETWORK_EPOCH_DURATION
```
NindoK, line 282: Use now as timestamp

script/DeployTanssiEcosystem.s.sol
```diff
@@ -274,31 +276,39 @@ contract DeployTanssiEcosystem is Script {
...
         );
-        _registerEntitiesToMiddleware();
-        networkMiddlewareService.setMiddleware(address(ecosystemEntities.middleware));
 
         ODefaultOperatorRewards operatorRewards =
```
NindoK, line 285: Deploy using deployRewards

src/contracts/middleware/Middleware.sol
```diff
@@ -539,4 +534,18 @@ contract Middleware is
             totalStake += operatorStake;
         }
     }
+
+    /**
+     * @inheritdoc SharedVaults
+     */
+    function _beforeRegisterSharedVault(address sharedVault, bytes memory data) internal virtual override {
```
NindoK, line 541: Put it before view functions. Follow the order external => public => internal => private => view function with same order

src/contracts/middleware/Middleware.sol
```diff
@@ -539,4 +534,18 @@ contract Middleware is
...
+        // TODO Steven: Cannot be empty
+        if (data.length == 0) {
+            return;
```
NindoK, line 544: Probably `revert Middleware__StakerInitParamsCantBeEmpty()`?

src/contracts/middleware/Middleware.sol
NindoK, line 541: Where are you overriding `_beforeRegisterSharedVault`? I don't see our own implementation

src/contracts/rewarder/ODefaultStakerRewardsFactory.sol
```diff
@@ -29,9 +29,6 @@ contract ODefaultStakerRewardsFactory is Registry, IODefaultStakerRewardsFactory
     ) {
         STAKER_REWARDS_IMPLEMENTATION = stakerRewardsImplementation;
     }
-    /**
-     * @inheritdoc IODefaultStakerRewardsFactory
```
NindoK, line 33: Why?

src/interfaces/middleware/IMiddleware.sol
```diff
@@ -22,6 +22,12 @@ interface IMiddleware {
      */
     event OperatorRewardContractSet(address indexed operatorRewardsAddress);
 
+    /**
+     * @notice Emitted when staker rewards factory is set
+     * @param stakerRewardsFactory Address of the staker rewards factory
+     */
+    event StakerRewardsFactorySet(address indexed stakerRewardsFactory);
```
NindoK, line 29: If I remember correctly we don't need this anymore right?

test/integration/Middleware.t.sol
```diff
@@ -1180,4 +1206,72 @@ contract MiddlewareTest is Test {
...
     }
+
+    // TODO: Change to check it fails registering shared vault if operator rewards is not set
+    function testWhenOperatorRewardsIsNotSetThenStakerRewardsAreNotDeployed() public {
+        vm.startPrank(owner);
+        // middleware.setOperatorRewardsContract(address(0)); // No need since it is currently set to 0. Also, not possible to set to 0.
+        VaultAddresses memory testVaultAddresses = _createTestVault(owner);
+        uint256 totalEntities = stakerRewardsFactory.totalEntities();
+
+        middleware.registerSharedVault(testVaultAddresses.vault);
```
NindoK, line 1217: Shouldnt this revert if no operatorRewards is not set?

test/integration/Middleware.t.sol
```diff
@@ -1180,4 +1206,72 @@ contract MiddlewareTest is Test {
...
+        vm.stopPrank();
+
+        // Since there is no operator rewards set, we cannot check stake rewards for vault, but we can check that no staker rewards are deployed:
+        assertEq(stakerRewardsFactory.totalEntities(), totalEntities);
+    }
+
+    function testWhenOperatorRewardsAndStakerRewardsFactoryAreSetWithVaultMissingStakerRewardsThenStakerRewardsAreDeployed(
```
NindoK, line 1224: As name use a shorter one since this is the base case. 
`testStakerRewardsIsDeployedWhenVaultIsRegistered`

demos/Demo.s.sol
```diff
@@ -245,9 +245,21 @@ contract Demo is Script {
         vm.startBroadcast(ownerPrivateKey);
         middleware.registerOperator(operator, OPERATOR_KEY, address(0));
         middleware.registerOperator(operator2, OPERATOR_KEY2, address(0));
-        middleware.registerSharedVault(address(vault));
-        middleware.registerSharedVault(address(vaultSlashable));
-        middleware.registerSharedVault(address(vaultVetoed));
+
+        bytes memory stakerRewardsData = abi.encode(
```
NindoK, line 249: Now that I look at it I feel we should just go ahead with the params in the struct instead of the bytes. I feel it makes the most sense

test/fork/Middleware.t.sol
```diff
@@ -708,11 +717,20 @@ contract MiddlewareTest is Test {
...
+        IODefaultStakerRewards.InitParams memory stakerRewardsParams = IODefaultStakerRewards.InitParams({
+            vault: address(0),
+            adminFee: 0,
+            defaultAdminRoleHolder: tanssi,
+            adminFeeClaimRoleHolder: address(0),
+            adminFeeSetRoleHolder: tanssi,
+            operatorRewardsRoleHolder: address(0),
+            network: tanssi
+        });
+        middleware2.registerSharedVault(address(ecosystemEntities.vault), stakerRewardsParams);
```
NindoK, line 733: How does this doesn't revert? We don't have a factory address set. Or at least it's only an address

test/integration/Rewards.t.sol
```diff
@@ -128,6 +132,8 @@ contract RewardsTest is Test {
     NetworkRegistry public networkRegistry;
     OptInService public operatorVaultOptInService;
     OptInService public operatorNetworkOptInService;
+    ODefaultOperatorRewards public operatorRewards;
+    uint48 public operatorShare = 2000;
```
NindoK, line 136: If it's a constant let's declare it constant and OPERATOR_SHARE

test/integration/Rewards.t.sol
```diff
@@ -245,7 +251,18 @@ contract RewardsTest is Test {
...
 
-        middleware = _deployMiddlewareWithProxy(tanssi, owner);
+        DeployRewards deployRewards = new DeployRewards();
+        address operatorRewardsAddress =
+            deployRewards.deployOperatorRewardsContract(tanssi, address(networkMiddlewareService), operatorShare);
+        operatorRewards = ODefaultOperatorRewards(operatorRewardsAddress);
+        address stakerRewardsFactoryAddress = makeAddr("stakerRewardsFactory");
+        middleware = _deployMiddlewareWithProxy(tanssi, owner, address(operatorRewards), stakerRewardsFactoryAddress);
+
+        vm.mockCall(
```
NindoK, line 261: Why do we need to mock it? This is integration test, we shouldn't mock if it's needed

src/contracts/rewarder/ODefaultStakerRewardsFactory.sol
steven2308: No idea :laughing:   Bringin it back

src/contracts/extensions/SharedVaults.sol
```diff
@@ -0,0 +1,90 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {BaseMiddleware} from "@symbiotic-middleware/middleware/BaseMiddleware.sol";
+import {ISharedVaults} from "../../interfaces/extensions/ISharedVaults.sol";
```
NindoK, line 5: Instead of using relative, use absolute with `src/interfaces` which starts directly from root and it's predictable

src/contracts/middleware/Middleware.sol
steven2308: It was on lib :sweat_smile:

src/contracts/middleware/Middleware.sol
```diff
@@ -539,4 +534,16 @@ contract Middleware is
...
     }
+
+    /**
+     * @inheritdoc SharedVaults
+     */
+    function _beforeRegisterSharedVault(
+        address sharedVault,
+        IODefaultStakerRewards.InitParams memory stakerRewardsParams
+    ) internal virtual override {
+        stakerRewardsParams.vault = sharedVault;
```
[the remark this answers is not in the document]
steven2308: Ok, but shall we at least check? That it matches the current vault?
I think it's important

src/contracts/middleware/Middleware.sol
steven2308: I moved this one, but many are misplaced, I think better to fix after merge to avoid conflicts

src/contracts/middleware/Middleware.sol
steven2308: It no longer applies since it's a struct now

src/contracts/middleware/Middleware.sol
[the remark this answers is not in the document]
NindoK: Yes you are right. If we check it's worse. Better like this

test/fork/Middleware.t.sol
steven2308: It does fail, problem is I was testing with coverage which excludes these tests.
Trying to fix now + deploying middleware using `DeployTanssiEcosystem`

steven2308: Now deploying real ones, but the setUp is broken (also on main), due to defaultColleteralAddress being 0x.

script/DeployTanssiEcosystem.s.sol
```diff
@@ -391,12 +411,23 @@ contract DeployTanssiEcosystem is Script {
...
+
+    function _initScriptsAndEntities(
+        HelperConfig _helperConfig
+    ) private {
+        contractScripts.helperConfig = _helperConfig;
+        contractScripts.deployVault = new DeployVault();
+        contractScripts.deployCollateral = new DeployCollateral();
+        contractScripts.deployRewards = new DeployRewards();
+
+        (address vaultConfiguratorAddress,,,,,,,,,) = _helperConfig.activeNetworkConfig();
```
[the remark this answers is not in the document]
steven2308: It is the reason why I did it. But all others are also needed inside the deploy so I didn't bother moving them.
However I could move them all into EcosystemEntity struct. Should I?

test/integration/Rewards.t.sol
steven2308: No longer mocked

test/integration/Middleware.t.sol
steven2308: There is no way for it to not be set

src/contracts/middleware/Middleware.sol
steven2308: The create method on the factory has some complex checks. I think it should be up to the factory (as it is now)

NindoK: Ok right! Forgot about it :D

script/DeployTanssiEcosystem.s.sol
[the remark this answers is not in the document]
NindoK: I think it makes most sense at this point.

test/unit/Middleware.t.sol
```diff
@@ -1729,9 +1720,6 @@ contract MiddlewareTest is Test {
         assertEq(middleware.VERSION(), 3);
         assertEq(MiddlewareV3(address(middleware)).i_operatorRewards(), address(operatorRewards));
 
-        vm.expectRevert(); //Doesn't exists
-        middleware.setOperatorRewardsContract(address(0));
```
NindoK, line 1783: Why?

test/unit/Middleware.t.sol
```diff
@@ -1712,14 +1708,9 @@ contract MiddlewareTest is Test {
     function testMiddlewareIsUpgradeableButMiddlewareV3IsNotUpgradeable() public {
         uint48 OPERATOR_SHARE = 2000;
 
-        ODefaultOperatorRewards operatorRewards =
-            new ODefaultOperatorRewards(tanssi, address(networkMiddlewareService), OPERATOR_SHARE);
-
         vm.prank(owner);
-        middleware.setOperatorRewardsContract(address(operatorRewards));
```
NindoK, line 1769: Change it with another function then that might exists. We just want to prove it that the upgrade works and storage remains the same. Probably use setGateway

NindoK (approved):

PR #44, opened 2025-03-06
Tests for upgrades and minor comments

NindoK:
- [ ] Added tests for upgrades. Both for normal upgrade and upgrade with to a non upgradeable contract. 
- [ ] Added some todos. 
- [ ] Now when operator is unregistered the key is also deleted.

### DIFF
--- a/test/unit/Middleware.t.sol
+++ b/test/unit/Middleware.t.sol
@@ -1862,7 +1862,6 @@ contract MiddlewareTest is Test {
         vm.warp(NETWORK_EPOCH_DURATION + 2);
         bytes memory key = middleware.getOperatorKeyAt(operator, uint48(block.timestamp));
 
-        // Assert key is zero since it's inactive
         assertEq(abi.decode(key, (bytes32)), OPERATOR_KEY);
     }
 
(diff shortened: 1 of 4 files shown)


test/unit/Middleware.t.sol
```diff
@@ -1684,4 +1718,241 @@ contract MiddlewareTest is Test {
...
+        middleware.registerOperator(operator, abi.encode(OPERATOR_KEY), address(0));
+
+        middleware.pauseOperator(operator);
+        vm.stopPrank();
+
+        vm.warp(NETWORK_EPOCH_DURATION + 2);
+        bytes memory key = middleware.getOperatorKeyAt(operator, uint48(block.timestamp));
+
+        // Assert key is zero since it's inactive
+        assertEq(abi.decode(key, (bytes32)), OPERATOR_KEY);
```
steven2308, line 1866: Should it be comparing to zero instead of `OPERATOR_KEY`?

test/unit/Middleware.t.sol
```diff
@@ -1684,4 +1718,241 @@ contract MiddlewareTest is Test {
...
+    function testGetOperatorKeyAtWithActivePreviousKey() public {
+        _registerOperatorToNetwork(operator, address(vault), false, false);
+
+        vm.startPrank(owner);
+        middleware.registerOperator(operator, abi.encode(OPERATOR_KEY), address(0));
+
+        uint48 activeKeyTimestamp = uint48(block.timestamp + 10);
+        vm.warp(activeKeyTimestamp);
+
+        middleware.updateOperatorKey(operator, abi.encode(PREV_OPERATOR_KEY));
```
steven2308, line 1878: I feel this should be `NEXT` instead of `PREV`.  The previous one is the one called `OPERATOR_KEY`.
Or am I getting it wrong?

NindoK: Ye changed earlier the things and forgot to change it thanks 😂

steven2308 (approved):

PR #45, opened 2025-03-10
[MD-943] - Added chainlink automation

NindoK:
This PR introduces Chainlink automation to send sorted operators' keys to Substrate via Snowbridge.


src/contracts/middleware/Middleware.sol
```diff
@@ -219,18 +263,54 @@ contract Middleware is
...
+    ) external view override returns (bool upkeepNeeded, bytes memory performData) {
+        uint48 epoch = getCurrentEpoch();
+        bytes32[] memory sortedKeys = sortOperatorsByVaults(epoch);
+
+        // Should it be epochDuration? Because we wanna send it once per network epoch
+        upkeepNeeded = (block.timestamp - s_lastTimestamp) > s_interval;
+
+        performData = abi.encode(sortedKeys, epoch);
+
+        return (upkeepNeeded, performData);
```
steven2308, line 290:
```suggestion

        return (upkeepNeeded, performData);
```
Not needed

src/contracts/middleware/Middleware.sol
```diff
@@ -219,18 +263,54 @@ contract Middleware is
...
+    ) external override onlyIfGatewayExists {
+        if (msg.sender != s_forwarderAddress) {
+            revert Middleware__NotForwarder();
+        }
+
+        if ((block.timestamp - s_lastTimestamp) > s_interval) {
+            s_lastTimestamp = block.timestamp;
+
+            // Decode the sorted keys and the epoch from performData
+            bytes32[] memory sortedKeys = abi.decode(performData, (bytes32[]));
```
steven2308, line 308: On the `checkUpKeep` you encode `keys + epoch`, but here you are only decoding the first. Is that ok?

steven2308 (approved):

steven2308 (approved):

nanocryk (approved):

steven2308 (commented):
Looks good, very minor comments.

test/unit/Middleware.t.sol
```diff
@@ -1644,7 +1654,7 @@ contract MiddlewareTest is Test {
     // ************************************************************************************************
 
     function testMiddlewareIsUpgradeable() public {
-        uint48 OPERATOR_SHARE = 2000;
+        // uint48 OPERATOR_SHARE = 2000;
```
steven2308, line 1657: Remove comment

steven2308 (commented):
Didn't we want to remove the `Middleware__AlreadySet` checks?

steven2308 (approved):

test/unit/Middleware.t.sol
NindoK: It's not needed for now. It's only needed if we correct the TODO comment below

src/contracts/middleware/Middleware.sol
```diff
@@ -226,9 +224,7 @@ contract Middleware is
     function setForwarder(
         address forwarder
     ) external checkAccess {
-        if (forwarder == address(0)) {
-            revert Middleware__InvalidAddress();
-        }
+        _checkNotZeroAddress(forwarder);
```
steven2308, line 227: Use the modifier.
I only used the internal method directly in the initialize because of stack too deep.
But actually now that it's a struct we might also use it there to further reduce the size.

PR #48, opened 2025-03-10
Adds mock to have full coverage on ShareVaults abstract.

steven2308:
Had to add a mock class so the default implementation of _beforeRegisterSharedVault was covered


src/contracts/extensions/SharedVaults.sol
```diff
@@ -19,7 +19,7 @@ abstract contract SharedVaults is BaseMiddleware, ISharedVaults {
     function registerSharedVault(
         address sharedVault,
         IODefaultStakerRewards.InitParams memory stakerRewardsParams
-    ) public checkAccess {
+    ) public virtual checkAccess {
```
NindoK, line 22: Why? We don't need this right? We don't want this to be virtual. Afaik we should just have a mock that doesn't set the _beforeRegisterSharedVaults correct?

test/unit/Middleware.t.sol
```diff
@@ -1896,4 +1897,15 @@ contract MiddlewareTest is Test {
 
         assertEq(abi.decode(key, (bytes32)), bytes32(0));
     }
+
+    // ************************************************************************************************
+    // *                                        Coverage
```
NindoK, line 1902: Put it under registerSharedVault

src/contracts/extensions/SharedVaults.sol
steven2308: Right, changed it a bit.

test/unit/utils/SharedVaultMock.sol
NindoK, line 1: Put it under mocks

NindoK (approved):

PR #52, opened 2025-03-12
[MD-929] StakerRewards are now upgradeable and use storage.

steven2308:


src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -155,9 +139,13 @@ contract ODefaultStakerRewards is
...
         }
 
+        __AccessControl_init();
+        __UUPSUpgradeable_init();
         __ReentrancyGuard_init();
-        VAULT = params.vault;
-        NETWORK = params.network;
+
+        StakerRewardsStorage storage $ = _getStakerRewardsStorage();
+        $.VAULT = params.vault;
```
NindoK, line 147: I'm thinking that probably these should be immutable or what you think? I mean once a staker reward is set the network and vault won't change. Actually if they would change it would be problematic innit?

NindoK: Especially since now dont have an implementation but we just deploy the contract.

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -272,15 +270,17 @@ contract ODefaultStakerRewards is
         bytes memory activeSharesHint,
         bytes memory activeStakeHint
     ) private {
-        if (_s_activeSharesCache[epoch] == 0) {
-            uint256 activeShares_ = IVault(VAULT).activeSharesAt(epochTs, activeSharesHint);
-            uint256 activeStake_ = IVault(VAULT).activeStakeAt(epochTs, activeStakeHint);
+        StakerRewardsStorage storage $ = _getStakerRewardsStorage();
```
NindoK, line 273: I feel we are pasing down the storage too many times. Wouldn't be better to just pass it down as input to the function?

NindoK: In all functions

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -314,54 +316,70 @@ contract ODefaultStakerRewards is
...
-        uint256 lastUnclaimedReward_ = s_lastUnclaimedReward[msg.sender][epoch][tokenAddress];
-
-        // Get the min between how many the user wants to claim and how many rewards are available
-        uint256 rewardsToClaim = Math.min(maxRewards, rewardsPerEpoch.length - lastUnclaimedReward_);
+        (uint256 rewardsToClaim, uint256 amount, uint256 lastUnclaimedReward_) =
+            _getRewardsToClaimAndAmount(epoch, tokenAddress, maxRewards, activeSharesOfHints);
 
         // If there are no rewards to claim, revert
         if (rewardsToClaim == 0) {
             revert ODefaultStakerRewards__NoRewardsToClaim();
```
NindoK, line 326: Put this directly in _getRewardsToClaim

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -314,54 +316,70 @@ contract ODefaultStakerRewards is
...
-        } else if (activeSharesOfHints.length != rewardsToClaim) {
-            revert ODefaultStakerRewards__InvalidHintsLength();
-        }
-
-        // Check the total amount for the user based on his shares in the vault and update the lastUnclaimedReward_
-        uint256 amount =
-            _claimRewardsPerEpoch(rewardsToClaim, rewardsPerEpoch, lastUnclaimedReward_, epoch, activeSharesOfHints);
-
-        s_lastUnclaimedReward[msg.sender][epoch][tokenAddress] = lastUnclaimedReward_;
+        $.lastUnclaimedReward[msg.sender][epoch][tokenAddress] = lastUnclaimedReward_ + rewardsToClaim;
```
NindoK, line 329: This update can be made dirctly inside `_getRewardsToClaimAndAmount`

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -314,54 +316,70 @@ contract ODefaultStakerRewards is
...
 
         // if the amount is greater than 0, transfer the tokens to the recipient
         if (amount > 0) {
             IERC20(tokenAddress).safeTransfer(recipient, amount);
         }
 
         emit ClaimRewards(
-            NETWORK, tokenAddress, msg.sender, epoch, recipient, lastUnclaimedReward_, rewardsToClaim, amount
+            $.NETWORK, tokenAddress, msg.sender, epoch, recipient, lastUnclaimedReward_, rewardsToClaim, amount
```
NindoK, line 337: lastUnclaimedReward_ should be lastUnclaimedReward_ + rewardsToClaim in this case

src/contracts/rewarder/ODefaultStakerRewards.sol
NindoK: So I would do: 

`lastUnclaimedReward_= lastUnclaimedReward_  + rewardsToClaim;`
`$.lastUnclaimedReward[msg.sender][epoch][tokenAddress] = lastUnclaimedReward_;`

NindoK: This is linked also to the comment below

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -407,13 +427,74 @@ contract ODefaultStakerRewards is
...
+     * @inheritdoc IODefaultStakerRewards
+     */
+    function claimableAdminFee(uint48 epoch, address tokenAddress) external view returns (uint256) {
+        StakerRewardsStorage storage $ = _getStakerRewardsStorage();
+        return $.claimableAdminFee[epoch][tokenAddress];
+    }
+
+    function _authorizeUpgrade(
+        address newImplementation
+    ) internal override onlyRole(DEFAULT_ADMIN_ROLE) {}
```
NindoK, line 480: 👍

test/unit/Rewards.t.sol
```diff
@@ -73,6 +73,9 @@ contract RewardsTest is Test {
     uint256 public constant ADMIN_FEE = 800; // 8%
     uint48 public constant OPERATOR_SHARE = 2000;
 
+    // keccak256(abi.encode(uint256(keccak256("tanssi.rewards.ODefaultStakerRewards.v1")) - 1)) & ~bytes32(uint256(0xff))
+    bytes32 constant MAIN_STORAGE_LOCATION = 0xe07cde22a6017f26eee680b6867ce6727151fb6097c75742cbe379265c377400;
```
NindoK, line 77: You can directly access this from a contract, so we dont need to maintain this

src/contracts/rewarder/ODefaultStakerRewards.sol
steven2308: I had this doubt, on the purpose of the event. But this behavior you mention makes more sense.

src/contracts/rewarder/ODefaultStakerRewards.sol
steven2308: All done.

src/contracts/rewarder/ODefaultStakerRewards.sol
steven2308: Passed for the 3 methods were it made sense. All others use it just in the same method.

steven2308: Nevermind, 2 of them produced stack too deep. I can try further tomorrow.

steven2308:
Addressed all the comments except passing down storage where possible. It only was possible in 3 places but 2 of them resulted in stack too deep. I can keep trying tomorrow but wouldn't block the merge and deploy for that.

Also agreed in making vault and network immutable instead. I do still include them in the init params since it's more convenient, but let me know your thoughts @NindoK

src/contracts/rewarder/ODefaultStakerRewards.sol
steven2308: Agree. Moved them.
I'd like to make all (EVERYWHERE) immutable uppercase, any objections? I think with that we can even remove the `i_` prefix.  I don't think we are storing any constant in storage (would it even make sense?) so we can easily tell them apart.

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -252,18 +274,24 @@ contract ODefaultStakerRewards is
...
 
-        // Check if the amount being sent is greater than 0
+        amount = _transferAndCheckAmount(tokenAddress, amount);
+
+        _updateAdminFeeAndRewards(amount, adminFee_, epoch, tokenAddress);
+
+        emit DistributeRewards(i_network, tokenAddress, eraIndex, epoch, amount, data);
+    }
+
+    function _transferAndCheckAmount(address tokenAddress, uint256 amount) private returns (uint256) {
```
NindoK, line 284: Just add here the named returns?

src/contracts/rewarder/ODefaultStakerRewards.sol
NindoK: Ye it's just that it feels like a waste of gas. If it's possible fine otherwise whatever

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -272,15 +300,17 @@ contract ODefaultStakerRewards is
         bytes memory activeSharesHint,
         bytes memory activeStakeHint
```
NindoK, line 301: Here shouldn't cause stack too deep right?

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -290,10 +320,12 @@ contract ODefaultStakerRewards is
         // And distribute the rest to the stakers
```
NindoK, line 320: Same here

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -334,34 +382,34 @@ contract ODefaultStakerRewards is
...
+            _claimRewardsPerEpoch($, rewardsToClaim, rewardsPerEpoch, lastUnclaimedReward_, epoch, activeSharesOfHints);
 
-        s_lastUnclaimedReward[msg.sender][epoch][tokenAddress] = lastUnclaimedReward_;
-
-        // if the amount is greater than 0, transfer the tokens to the recipient
-        if (amount > 0) {
-            IERC20(tokenAddress).safeTransfer(recipient, amount);
+        // If there are no rewards to claim, revert
+        if (rewardsToClaim == 0) {
+            revert ODefaultStakerRewards__NoRewardsToClaim();
```
NindoK, line 390: Should go after 376

src/contracts/rewarder/ODefaultStakerRewards.sol
steven2308: It causes the stack too deep on the caller. I tried again and discovered I can do one of them but not 2. :man_shrugging: 
This and the next one are called from the same method.

src/contracts/rewarder/ODefaultStakerRewards.sol
steven2308: It causes the stack too deep on the caller.

src/contracts/rewarder/ODefaultStakerRewards.sol
steven2308: I'll do this in another PR.

src/contracts/rewarder/ODefaultStakerRewards.sol
NindoK: Whaaaaattt 😨

src/contracts/rewarder/ODefaultStakerRewards.sol
NindoK: I'd keep the i_ for being consistent across all other contracts

NindoK (approved):

PR #57, opened 2025-03-14
[MD-944] Add Chainlink to get price of the underlying registered vault

steven2308:


src/contracts/extensions/OSharedVaults.sol
```diff
@@ -22,6 +22,7 @@ abstract contract OSharedVaults is BaseMiddleware, IOSharedVaults {
     ) public checkAccess {
         _beforeRegisterSharedVault(sharedVault, stakerRewardsParams);
         _registerSharedVault(sharedVault);
+        _afterRegisterSharedVault(sharedVault);
```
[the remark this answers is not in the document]
steven2308: I'm using this hook to map vault to collateral. To do that I need to call `IVault(vault).collateral()` or similar.
If I do that on the the before hook, it might be that the vault is not a valid vault so it doesn't have the collateral method, which produces an unexpected revert.
If I do it after, there's no such issue, since the `_register` method checks for its validity.  So there were 3 options:
1. Assume the vault is good. (I don't like assuming)
2. Check the vault is valid on the `_before` hook or when getting the collateral. (Would do double check).
3. Create and after hook. I went for this.

Now that I think of it I could also have a sort of try catch when getting the collateral, but feels like a weird practice in Solidity IMO.

src/contracts/middleware/Middleware.sol
```diff
@@ -247,15 +286,15 @@ contract Middleware is
     //TODO use prebuilt function from SDK
     function slash(
         uint48 epoch,
-        bytes32 operatorKey,
+        bytes32 operatorKey_,
```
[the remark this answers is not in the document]
steven2308: It was shadowing the state variable.

src/contracts/extensions/OSharedVaults.sol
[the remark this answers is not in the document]
NindoK: Actually no it totally makes sense. But at this point I would also move what we have in the beforeRegister to the after. Since "theoretically" we want to deploy the contract once it has been validated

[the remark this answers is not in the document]
steven2308: Moved to after hook.

src/contracts/middleware/Middleware.sol
```diff
@@ -65,18 +71,24 @@ contract Middleware is
     using Subnetwork for address;
     using Math for uint256;
 
+    modifier notZeroAddress(
```
NindoK, line 74: Not a fan of this. I don't think it optimize the code and it increases code length

src/contracts/middleware/Middleware.sol
```diff
@@ -153,6 +177,18 @@ contract Middleware is
         _grantRole(GATEWAY_ROLE, _gateway);
     }
 
+    function setCollateralToOracle(
+        address collateral,
+        address oracle
+    ) external checkAccess notZeroAddress(collateral) {
```
NindoK, line 183: Do we need here the check on zero address? Probably not right?

test/fork/Middleware.t.sol
```diff
@@ -494,175 +507,115 @@ contract MiddlewareTest is Test {
...
+        (uint48 currentEpoch, Middleware.ValidatorData[] memory validators,,,, uint256 powerFromSharesOperator3) =
+            _prepareSlashingTest();
 
-        //We calculate the amount slashable for only the operator3 since it's the only one that should be slashed. As a side effect operator2 will be slashed too since it's taking part in a NetworkRestake delegator based vault
-        uint256 slashAmountSlashable3 = (SLASH_AMOUNT * remainingOperator3Stake) / totalOperator3Stake;
-        uint256 slashedAmount = 30 ether;
-        // We want to slash 30 ether, so we need to calculate what percentage
-        uint256 slashingFraction = slashedAmount.mulDiv(PARTS_PER_BILLION, totalOperator3Stake);
+        // We only take half of the operator3 shares, since only its participation on vaultSlashable will be slashed, regular vault isn't affected
+        uint256 slashingPower = (SLASHING_FRACTION * (powerFromSharesOperator3 / 2)) / PARTS_PER_BILLION;
```
NindoK, line 527: Why 10% has been used instead of the slshedAmount?

test/integration/Middleware.t.sol
```diff
@@ -107,6 +114,17 @@ contract MiddlewareTest is Test {
     uint256 public constant OPERATOR_SHARE = 1;
     uint256 public constant TOTAL_NETWORK_SHARES = 3;
     uint256 public constant PARTS_PER_BILLION = 1_000_000_000;
+    uint256 public constant SLASHING_FRACTION = PARTS_PER_BILLION / 10; // 10%
+
+    // TODO Steven: Change to realistic values
+    uint8 public constant ORACLE_DECIMALS = 10; // TODO Steven: This decimals are actually on staked value, not on price
```
NindoK, line 120: Why not standard 18?

test/integration/Middleware.t.sol
```diff
@@ -362,11 +388,24 @@ contract MiddlewareTest is Test {
         vm.stopPrank();
     }
 
-    function _registerOperator(address _operator, address _network, address _vault) public {
+    function _registerOperatorAndOptIn(
+        address _operator,
+        address _network,
+        address _vault,
+        address vault2,
+        address vault3
```
NindoK, line 396: Wouldn't it be better to have a boolean is that conditionally registerOperator and optIn to the network, so we can reuse the same function for whatever vault. Otherwise if we will need 4 vaults we will have to extend this

src/contracts/middleware/Middleware.sol
steven2308: This actually reduced the contract enough to fit.  The trick is using an internal function in the modifier.

src/contracts/middleware/Middleware.sol
steven2308: It wouldn't make sense to set zero address to an oracle. I think the check is important for collateral.
For oracle, I'm not checking so you can set the oracle of a collateral to zero, in case you don't want to support it anymore.

test/fork/Middleware.t.sol
steven2308: Previously tests were using an amount to calculate the fraction (10% here).  I didn't see the point to start from amount if you need % anyway.

test/integration/Middleware.t.sol
steven2308: I like it, will do.

src/contracts/middleware/Middleware.sol
NindoK: Yes it doesn't make sense but I don't think it's an issue per se. By taking this out we are taking out a bit of code too!

src/contracts/middleware/Middleware.sol
NindoK: Oh cool. Wouldn't just the if be the same if not shorter? Or it's because with an internal function there is a pointer and thus you don't copy paste the code every time the modifier is used?

steven2308: Exactly, that's the trick.

src/interfaces/extensions/IOERC20.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+/**
+ * @title IOERC20
+ * @notice Interface for managing ERC20 tokens, including decimals
+ */
+
+interface IOERC20 is IERC20 {
```
NindoK, line 10: Use IERC20Metadata from OZ

src/contracts/middleware/Middleware.sol
```diff
@@ -14,19 +14,26 @@
...
 
+//**************************************************************************************************
+//                                      CHAINLINK
+//**************************************************************************************************
+import {AggregatorV3Interface} from "@chainlink/shared/interfaces/AggregatorV2V3Interface.sol";
+
 //**************************************************************************************************
 //                                      OPENZEPPELIN
 //**************************************************************************************************
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
NindoK, line 25: You can take this out now right?

NindoK: You are right. Dunno why it didn't popped up on my browser :/

src/contracts/middleware/Middleware.sol
```diff
@@ -126,19 +141,34 @@ contract Middleware is
...
+        address collateral = vaultToCollateral(vault);
+        address oracle = collateralToOracle(collateral);
+
+        if (oracle == address(0)) {
+            revert Middleware__NotSupportedCollateral(collateral);
+        }
+        (, int256 price,,,) = AggregatorV3Interface(oracle).latestRoundData();
+        uint8 priceDecimals = AggregatorV3Interface(oracle).decimals();
+        power = stake.mulDiv(uint256(price), 10 ** priceDecimals);
+        // Normalize power to 18 decimals
```
NindoK, line 167: I'm still thinking that this is wrong. We need a better formula, but we can do it in another PR

NindoK (approved):

PR #64, opened 2025-03-27
deploy tanssi ecosystem

girazoki:
adapt to more complex scenario in which operators are registered in different vaults


demos/DeployTanssiEcosystemDemo.s.sol
```diff
@@ -290,9 +290,16 @@ contract DeployTanssiEcosystem is Script {
         tokensAddresses.stETHToken.transfer{gas: 1_000_000}(operator, 1000 ether);
         tokensAddresses.stETHToken.transfer{gas: 1_000_000}(operator2, 1000 ether);
         tokensAddresses.stETHToken.transfer{gas: 1_000_000}(operator3, 1000 ether);
```
steven2308, line 292: I think this one is no longer needed

steven2308 (approved):
Besides the possibly unneeded transfer, all good.

demos/DeployTanssiEcosystemDemo.s.sol
steven2308: No, because operators will use those to deposit to vaults.

PR #65, opened 2025-03-28
Shrinked Middleware size

NindoK:
26k kb => 23476 with 1k optimization.

With 200 we get around 19k KB


src/contracts/middleware/Middleware.sol
```diff
@@ -411,7 +411,7 @@ contract Middleware is
             return;
         }
         uint256 slasherType = IEntity(slasher).TYPE();
-        //? Shall we add event for each slash?
+        //TODO? Shall we add event for each slash?
```
steven2308, line 414: For me it is a definite yes. Not in this PR ofc.

test/unit/Middleware.t.sol
```diff
@@ -871,29 +884,29 @@ contract MiddlewareTest is Test {
...
+    //     vault.setSlasher(address(slasher));
+    //     vm.store(address(slasher), bytes32(uint256(0)), bytes32(uint256(uint160(address(vault)))));
+    //     middleware.registerSharedVault(address(vault), stakerRewardsParams);
+    //     vm.warp(START_TIME + SLASHING_WINDOW + 1);
+    //     uint48 currentEpoch = middleware.getCurrentEpoch();
+    //     vm.warp(START_TIME + SLASHING_WINDOW - 1);
+    //     vm.expectRevert(IMiddleware.Middleware__InvalidEpoch.selector);
+    //     OBaseMiddlewareReader(address(middleware)).getTotalStake(currentEpoch + 1);
+    //     vm.stopPrank();
+    // }
```
steven2308, line 909: What's up with this test?

steven2308 (commented):
Looks good but can you remind me why we couldn't extend the BaseMiddleware? Why did we have to copy it. Are you making some optimizations maybe? If so, which?

I tested what we talked yesterday, an unused internal method does not affect contract size.

steven2308:
Tests are breaking due to a bad import also.

test/unit/Middleware.t.sol
NindoK: We don't have anymore the revert. Since it's a view function and we don't want that

steven2308: So let's just remove, no?  Why leave it commented?

steven2308 (approved):

PR #66, opened 2025-03-31
Added events for setter

NindoK:


src/interfaces/middleware/IMiddleware.sol
```diff
@@ -23,6 +23,24 @@ interface IMiddleware {
      */
     event CollateralToOracleSet(address indexed collateral, address indexed oracle);
 
+    /**
+     * @notice Emitted when the interval for which the `performUpkeep` should be performed.
```
steven2308, line 27: This is a bit off.
```suggestion
     * @notice Emitted when the interval for which the `performUpkeep` should be performed is set.
```

steven2308 (approved):
Minor details.
I suggest we merge the size reduction PR first

PR #67, opened 2025-03-31
[MD-973] Staker rewards distribution based on operator power

steven2308:


src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -196,27 +200,93 @@ contract ODefaultOperatorRewards is
...
+            uint256 amountForVault;
+            // Last vault gets the remaining amount
+            if (i == totalVaults - 1) {
+                amountForVault = stakerAmount - distributedAmount;
+            } else {
+                amountForVault = vaultPowers[i].mulDiv(stakerAmount, totalPower);
+            }
+            amountPerVault[i] = amountForVault;
+            unchecked {
+                distributedAmount += amountForVault;
```
NindoK, line 261: Hypothetically this shouldn't overflow right? i'm just thinking if there's an attack vector

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -196,27 +200,93 @@ contract ODefaultOperatorRewards is
...
+        (, address[] memory operatorVaults) = Middleware(middlewareAddress).getOperatorVaults(operator, epochStartTs);
 
-        // TODO: Currently it's only for a specific vault. We don't care now about making it able to send rewards for multiple vaults. It's hardcoded to the first vault of the operator.
+        // The operation is divided into 3 steps which repeat the for loop over each vault.
+        // It was necessary to avoid stack too deep errors due to the high number of variables
+        uint256 totalVaults = operatorVaults.length;
+        (uint256[] memory vaultPowers, uint256 totalPower) =
+            _getOperatorPowerPerVault(operatorVaults, totalVaults, epochStartTs, operator, middlewareAddress);
+
+        uint256[] memory amountPerVault = _getRewardsAmountPerVault(stakerAmount, totalVaults, vaultPowers, totalPower);
```
NindoK, line 218: Can you create a test to have a benchmark for gas? I'm really afraid of this function xD

test/unit/Rewards.t.sol
```diff
@@ -456,6 +470,79 @@ contract RewardsTest is Test {
...
+
+        vm.mockCall(
+            address(middleware),
+            abi.encodeWithSelector(IMiddleware.getOperatorVaults.selector, alice, epochStartTs),
+            abi.encode(2, vaults)
+        );
+
+        // Selector is for IBaseMiddlewareReader.getOperatorPowerAt(uint48,address,address,uint96) returns (uint256). The method has 3 implementations
+        vm.mockCall(
+            address(middleware), abi.encodeWithSelector(0xbcdc52be, epochStartTs, alice, vaults[0], 0), abi.encode(40)
```
NindoK, line 493: Let's do it like this, so we can easily modify it
`bytes4 selector = bytes4(keccak256("getOperatorPowerAt(uint48,address,address,uint96)"));`

test/unit/Rewards.t.sol
```diff
@@ -456,6 +470,79 @@ contract RewardsTest is Test {
...
+        emit IODefaultStakerRewards.DistributeRewards(
+            tanssi, address(token), eraIndex, epoch, expectedAmountVault1, REWARDS_ADDITIONAL_DATA
+        );
+
+        vm.expectEmit(true, true, false, true);
+        emit IODefaultOperatorRewards.ClaimRewards(
+            recipient, address(token), eraIndex, epoch, address(this), EXPECTED_CLAIMABLE
+        );
+        operatorRewards.claimRewards(claimRewardsData);
+
```
NindoK, line 541: We need to check that the amount is in that contract too. And since there are 2 vaults we should also check how much is going to the other vault right?

Let's use balanceOf to see how much the contract has

test/unit/Rewards.t.sol
```diff
@@ -456,6 +470,79 @@ contract RewardsTest is Test {
         assertEq(amountClaimed_, EXPECTED_CLAIMABLE);
     }
 
+    function testClaimRewardsWithMultipleVaults() public {
```
NindoK, line 473: Let's make some integrations tests too!

src/contracts/rewarder/ODefaultOperatorRewards.sol
steven2308: The max value for this will be the rewards amount, which was already fitting in `uint256`

test/unit/Rewards.t.sol
steven2308: I knew there was a better way, can't believe I missed it :laughing:

NindoK: 😂

src/contracts/rewarder/ODefaultOperatorRewards.sol
steven2308: On integration tests: 
I simply checked the gas consumption on the integration test and logged it, I don't think we need a separate test.

For an operator using 3 vaults the claim process takes:
`Total gas used:  856420`
Seems well below the limit. Do we have an estimation of the number a vaults operators will have?

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -197,24 +202,85 @@ contract ODefaultOperatorRewards is
...
-        //TODO: For now this is expected to be a single vault. Change it to be able to handle multiple vaults.
         (, address[] memory operatorVaults) =
-            IOBaseMiddlewareReader(middlewareAddress).getOperatorVaults(recipient, epochStartTs);
+            IOBaseMiddlewareReader(middlewareAddress).getOperatorVaults(operator, epochStartTs);
+
+        uint256 totalVaults = operatorVaults.length;
+        uint256[] memory amountPerVault = _getRewardsAmountPerVault(
+            operatorVaults, totalVaults, epochStartTs, operator, middlewareAddress, stakerAmount
+        );
+
```
NindoK, line 219: We should check that `amountPerVault.length > 0` right? Otherwise we should revert

NindoK: Or totalVaults. Both are fine

test/integration/Rewards.t.sol
```diff
@@ -84,20 +90,33 @@ contract RewardsTest is Test {
...
-    bytes32 public constant OPERATOR2_KEY = bytes32(uint256(2));
-    bytes32 public constant OPERATOR3_KEY = bytes32(uint256(3));
-    uint48 public constant OPERATOR_SHARE = 2000;
-    uint256 public constant TOTAL_NETWORK_SHARES = 3;
-    uint256 public constant PARTS_PER_BILLION = 1_000_000_000;
+    uint48 public constant OPERATOR_SHARE = 2000; // 20%
     uint256 public constant ONE_DAY = 86_400;
 
+    // We use pregenerated keys and proofs for the operators, alice becomes operator 3
+    bytes32 public OPERATOR_KEY = bytes32(uint256(1));
```
NindoK, line 98: Make it constant. The others are not anymore constant though right?

test/integration/Rewards.t.sol
```diff
@@ -589,4 +627,106 @@ contract RewardsTest is Test {
...
+        );
+        uint256 powerVault3 = IOBaseMiddlewareReader(address(middleware)).getOperatorPowerAt(
+            epochStartTs, operator_, address(vaultVetoed), uint96(0)
+        );
+
+        uint256 totalPower = powerVault1 + powerVault2 + powerVault3;
+        rewards = new uint256[](3);
+
+        rewards[0] = (expectedAmountStakers * powerVault1) / totalPower;
+        rewards[1] = (expectedAmountStakers * powerVault2) / totalPower;
```
NindoK, line 718: Use `mulDiv` to be consistent with Math

src/contracts/rewarder/ODefaultOperatorRewards.sol
steven2308: Added. This will prevent the operator from claiming his rewards if he has no vaults, which I actually like. It's a good protection for stakers.
As it was before your suggestions, if for some reason the operator has no vaults we would still be able to get a chunk and leave nothing for stakers.

NindoK (approved):

src/contracts/rewarder/ODefaultOperatorRewards.sol
NindoK: Looks nice. It's not such a big number

PR #68, opened 2025-03-31
[MD-929] Adds missing address-zero checks

steven2308:
Checks are added on constructor and initializer.


NindoK (approved):

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -221,10 +231,10 @@ contract ODefaultOperatorRewards is
...
      * @inheritdoc IODefaultOperatorRewards
      */
-    function setStakerRewardContract(address stakerRewards, address vault) external onlyMiddleware {
-        if (stakerRewards == address(0) || vault == address(0)) {
-            revert ODefaultOperatorRewards__InvalidAddress();
-        }
+    function setStakerRewardContract(
+        address stakerRewards,
+        address vault
+    ) external onlyMiddleware notZeroAddress(stakerRewards) notZeroAddress(vault) {
```
NindoK, line 237: The only thing that I don't like is that this is not veri optimised. Because with the or it should be better right?

steven2308: It could be slightly better on gas, but this happens on operations which occur just once or very rarely.  I think it's not a reason to worry.

PR #69, opened 2025-04-03
[MD-987] - Fixed total operator shares and and operator limit for full restake

NindoK:
This PR fixes:

- attribution of operator shares and operator network limits, respectively for `NetworkRestake` and `FullRestake`, to be congruent with their partecipation in vaults.
- Minor fixes with unused variables and added constant for decimals

This PR also assess the soundness of math for `stakeToPower`.

### DIFF
--- a/src/contracts/middleware/Middleware.sol
+++ b/src/contracts/middleware/Middleware.sol
@@ -161,22 +161,6 @@ contract Middleware is
         address newImplementation
     ) internal override checkAccess {}
 
-    /**
-     * @inheritdoc OSharedVaults
-     */
-    function _afterRegisterSharedVault(
-        address sharedVault,
-        IODefaultStakerRewards.InitParams memory stakerRewardsParams
-    ) internal override {
-        address stakerRewards =
-            IODefaultStakerRewardsFactory(i_stakerRewardsFactory).create(sharedVault, stakerRewardsParams);
-
-        IODefaultOperatorRewards(i_operatorRewards).setStakerRewardContract(stakerRewards, sharedVault);
-
-        address collateral = IVault(sharedVault).collateral();
-        _setVaultToCollateral(sharedVault, collateral);
-    }
-
     function stakeToPower(address vault, uint256 stake) public view override returns (uint256 power) {
         address collateral = vaultToCollateral(vault);
         address oracle = collateralToOracle(collateral);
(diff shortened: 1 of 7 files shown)


src/contracts/middleware/Middleware.sol
```diff
@@ -189,8 +173,8 @@ contract Middleware is
         power = stake.mulDiv(uint256(price), 10 ** priceDecimals);
         // Normalize power to 18 decimals
         uint8 collateralDecimals = IERC20Metadata(collateral).decimals();
-        if (collateralDecimals != uint8(18)) {
-            power = power.mulDiv(10 ** 18, 10 ** collateralDecimals);
+        if (collateralDecimals != DEFAULT_DECIMALS) {
```
steven2308, line 176: Nice one

test/fork/Middleware.t.sol
```diff
@@ -245,7 +245,7 @@ contract MiddlewareTest is Test {
         }
 
         {
-            totalFullRestakePower = ((OPERATOR_STAKE * 2) * uint256(ORACLE_CONVERSION_TOKEN)) / 10 ** ORACLE_DECIMALS;
+            totalFullRestakePower = (OPERATOR_STAKE * uint256(ORACLE_CONVERSION_TOKEN)) / 10 ** ORACLE_DECIMALS;
```
steven2308, line 248: I'm confused here. Operator 2 and 3 are depositing on the full restake vault, it should be `*2`

test/fork/Middleware.t.sol
```diff
@@ -461,9 +452,6 @@ contract MiddlewareTest is Test {
             _prepareSlashingTest();
 
         //Since vaultVetoed is full restake, it exactly gets the amount deposited, so no need to calculations
-        uint256 activeStakeInVetoed = ecosystemEntities.vaultVetoed.activeStake();
-        uint256 activePowerInVetoed = (activeStakeInVetoed * uint256(ORACLE_CONVERSION_TOKEN)) / 10 ** ORACLE_DECIMALS;
-        assertEq(activePowerInVetoed, totalFullRestakePower);
```
steven2308, line 466: Why remove this check? 
You are even replacing one for the other on the test, I think it's valuable to have this assertion.

steven2308 (commented):
A lot of good improvements, but I'm concerned about the change on `totalFullRestakePower`. I'm probably missing something.

Also, calculations on middleware were fine, right?

test/fork/Middleware.t.sol
NindoK: No because when you set maxOperatorShare for full restake you basically cap their stake! That's why my changes

NindoK: With the limits we put there they could had whatever amount as their stake. Hence the increased security risk for FullRestake

test/fork/Middleware.t.sol
NindoK: Because it relates to the previous comment. Now they won't be equal

test/fork/Middleware.t.sol
steven2308: Crystal clear

steven2308 (approved):

PR #70, opened 2025-04-04
[MD-1003] Send sorted operators to substrate

steven2308:
In `performUpkeep` and `sendCurrentOperatorsKeys` we were already getting operators sorted by power.
However there was a proper test missing. I added it in integration tests where there is a better environment configured to asses the order is correct and that it changes when power changes. On unit tests I gave it a tried but I had to mock so many things that it didn't really make sense.

Renamed `sortOperatorsByVaults` to `sortOperatorsByPower`, which seems a clearer name for me. I checked and this method does not exist on the SDK so I'm not breaking any compatiblity, this is just for us.

Also added a comment on `sendCurrentOperatorsKeys`, I think the method should be gated since it calls the gateway, right now anyone can call it.

### DIFF
--- a/src/contracts/middleware/Middleware.sol
+++ b/src/contracts/middleware/Middleware.sol
@@ -283,7 +283,6 @@ contract Middleware is
      * @inheritdoc IMiddleware
      */
     function sendCurrentOperatorsKeys() external returns (bytes32[] memory sortedKeys) {
-        // TODO: How come this is not gated?
         address gateway = getGateway();
         if (gateway == address(0)) {
             revert Middleware__GatewayNotSet();
--- a/src/contracts/middleware/Middleware.sol
+++ b/src/contracts/middleware/Middleware.sol
@@ -289,7 +289,7 @@ contract Middleware is
         }
 
         uint48 epoch = getCurrentEpoch();
-        sortedKeys = IOBaseMiddlewareReader(address(this)).sortOperatorsByVaults(epoch);
+        sortedKeys = IOBaseMiddlewareReader(address(this)).sortOperatorsByPower(epoch);
         IOGateway(gateway).sendOperatorsData(sortedKeys, epoch);
     }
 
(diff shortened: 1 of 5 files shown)


test/unit/Middleware.t.sol
```diff
@@ -2440,7 +2439,7 @@ contract MiddlewareTest is Test {
 
     function _prepareInitializeTest() private returns (Middleware middleware2, IMiddleware.InitParams memory params) {
         middleware2 = Middleware(address(new MiddlewareProxy(address(middlewareImpl), "")));
-        address readHelper = address(new OBaseMiddlewareReader());
+        address readHelper_ = address(new OBaseMiddlewareReader());
```
steven2308, line 2442: Previously it was ghosting another definition.

src/contracts/middleware/Middleware.sol
```diff
@@ -283,27 +283,28 @@ contract Middleware is
      * @inheritdoc IMiddleware
      */
     function sendCurrentOperatorsKeys() external returns (bytes32[] memory sortedKeys) {
+        // TODO: How come this is not gated?
```
NindoK, line 286: Because everyone can call it since it cost money (quite a lot) and doesn't affect us in any way

test/integration/Middleware.t.sol
```diff
@@ -1343,4 +1342,38 @@ contract MiddlewareTest is Test {
...
+        IFullRestakeDelegator(vaultAddresses.delegatorVetoed).setOperatorNetworkLimit(
+            tanssi.subnetwork(0), operator3, 0
+        );
+
+        vm.warp(block.timestamp + VAULT_EPOCH_DURATION + 1);
+
+        keys = middleware.sendCurrentOperatorsKeys();
+        assertEq(keys.length, 3);
+
+        // Now OP2 > OP3 > OP1 (In terms of power)
```
NindoK, line 1374: How come op3 > op1? Op1 has stake while op3 is set to 0 right?

test/unit/Middleware.t.sol
NindoK: Could just override the variable

test/integration/Middleware.t.sol
steven2308: Op 3 has stake in other 2 vaults. Op only in one of those.

test/unit/Middleware.t.sol
steven2308: That gives me a warning on the UI. I prefer not having warnings at all so when there is one I know I should care.

test/integration/Middleware.t.sol
NindoK: Oh right. Makes sense

PR #72, opened 2025-04-11
[MD-1009] Stronger integration tests

steven2308:


src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -139,7 +139,7 @@ contract ODefaultOperatorRewards is
             epoch: epoch,
             amount: amount,
             // We need to calculate how much each point is worth in tokens
-            tokensPerPoint: amount / totalPointsToken, // TODO: To change/check the math.
+            tokensPerPoint: amount / totalPoints, // TODO: To change/check the math. There will be a rounding error so some rewards will be forever stuck in this contract.
```
steven2308, line 142: We could do something like this to keep track of leftovers so we can later claim, or just for the sake of tracking them.
```
tokensPerPoint = amount / totalPoints
$.leftOvers += tokensPerPoint - amount*totalPoints
```
But that would imply either a change in storage + migration or at adding a new tiny second storage.

steven2308: @NindoK suggests we can also simply calculate it off chain by tracking DistributeRewards. We could do the same math since the event includes `amount` and `tokensPerPoint`

NindoK: Offchain. Don't wanna increase gas cost for 1 wei 😂

script/DeployVault.s.sol
```diff
@@ -54,6 +55,7 @@ contract DeployVault is Script {
         address vaultConfigurator;
         address collateral;
         address owner;
+        address operator;
     }
 
     enum DelegatorIndex {
```
NindoK, line 61: The new indexes are:

```
enum DelegatorType {
        NETWORK_RESTAKE,
        FULL_RESTAKE,
        OPERATOR_SPECIFIC,
        OPERATOR_NETWORK_SPECIFIC
    }
    ```
    
    I think we should update it

File: src/contracts/middleware/Middleware.sol [hunk 3]
```diff
@@ -340,7 +340,10 @@ contract Middleware is
         uint48 epochStartTs = IOBaseMiddlewareReader(address(this)).getEpochStart(epoch);
         address operator = operatorByKey(abi.encode(operatorKey));
 
-        // for epoch older than SLASHING_WINDOW total stake can be invalidated (use cache)
+        if (_SLASHING_WINDOW() > Time.timestamp()) {
```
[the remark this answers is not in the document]
steven2308: If you try slashing early the code reverts with panic due to underflow in 2 places. One was in middleware  (I inverted the logic so it did not revert), but the other one is in VaultManager or somewhere we do not overwrite. 

Reverting like this is no-no for me, so I added this check.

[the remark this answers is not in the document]
NindoK: You have underflow only because in tests you start at 1 timestamp. If you force the timestamp to be higher than slashing window this can't occur. I'd take this out as it would waste gas

script/DeployVault.s.sol
```diff
@@ -54,12 +55,14 @@ contract DeployVault is Script {
...
         NETWORK_RESTAKE, // 0
         FULL_RESTAKE, // 1
-        OPERATOR_SPECIFIC // 2
+        OPERATOR_SPECIFIC, // 2
+        OPERATOR_NETWORK_SPECIFIC // 3
```
NindoK, line 65: We should test this one too. In our case there might be in the future a possibility in which there will be a single operator for a specific network. So now that we have some free time it's a  nice test

test/unit/Deploy.t.sol
```diff
@@ -571,7 +571,8 @@ contract DeployTest is Test {
             shouldBroadcast: false,
             vaultConfigurator: address(0),
             collateral: stEth,
-            owner: tanssi
+            owner: tanssi,
+            operator: address(0)
```
NindoK, line 575: Let's keep tanssi for operator instead of address(0)

test/unit/Middleware.t.sol
```diff
@@ -1275,6 +1275,14 @@ contract MiddlewareTest is Test {
         middleware.slash(currentEpoch, OPERATOR_KEY, slashPercentage);
     }
 
+    function testSlashingTooEarly() public {
```
NindoK, line 1278: Remove based on the other comment

test/integration/Rewards.t.sol
```diff
@@ -380,7 +380,8 @@ contract RewardsTest is Test {
             shouldBroadcast: false,
             vaultConfigurator: address(vaultConfigurator),
             collateral: address(stETH),
-            owner: _owner
+            owner: _owner,
+            operator: address(0)
```
NindoK, line 384: Lett's keep tanssi as operator

test/integration/Middleware.t.sol
```diff
@@ -367,7 +367,8 @@ contract MiddlewareTest is Test {
             shouldBroadcast: false,
             vaultConfigurator: address(vaultConfigurator),
             collateral: address(stETH),
-            owner: _owner
+            owner: _owner,
+            operator: address(0)
```
NindoK, line 371: Lett's keep tanssi as operator

script/DeployVault.s.sol
```diff
@@ -234,7 +238,8 @@ contract DeployVault is Script {
             shouldBroadcast: true,
             vaultConfigurator: vaultConfigurator,
             collateral: collateral,
-            owner: owner
+            owner: owner,
+            operator: address(0)
```
NindoK, line 242: Lett's keep owner as operator

script/DeployVault.s.sol
NindoK: Also we can actually import it instead of maintaining it.

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+    uint48 public constant NETWORK_EPOCH_DURATION = 2 days;
+    uint48 public constant SLASHING_WINDOW = 5 days;
+    uint48 public constant VETO_DURATION = 1 days;
+
+    // Token and Oracle decimals + conversion rates
+    uint8 public constant TOKEN_DECIMALS_BTC = 18;
+    uint8 public constant TOKEN_DECIMALS_ETH = 18;
+    uint8 public constant TOKEN_DECIMALS_USDC = 6;
+
+    uint8 public constant ORACLE_DECIMALS_ETH = 2;
```
NindoK, line 90: Why these strange numbers for eth and btc. While usdc gets a full 8? xD

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+            adminFeeClaimRoleHolder: tanssi,
+            adminFeeSetRoleHolder: tanssi
+        });
+        middleware.registerSharedVault(address(vaultsData.v1.vault), stakerRewardsParams);
+        middleware.registerSharedVault(address(vaultsData.v2.vault), stakerRewardsParams);
+        middleware.registerSharedVault(address(vaultsData.v3.vault), stakerRewardsParams);
+        middleware.registerSharedVault(address(vaultsData.v4.vault), stakerRewardsParams);
+        middleware.registerSharedVault(address(vaultsData.v5.vault), stakerRewardsParams);
+
+        middleware.registerOperator(operator1, abi.encode(OPERATOR1_KEY), address(0));
```
NindoK, line 473: when registering an operator, a vault should be passed to make the `registerOperator` to register an operator vault!

NindoK: All that flow should be tested!

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+        // On Vault 4: delegator is network restake and OP5 is assigned 2/3 shares
+        uint256 expectedOperatorPower5 = VAULT3_TOTAL_STAKE.mulDiv(OPERATOR5_SHARES_V3, VAULT3_TOTAL_SHARES).mulDiv(
+            uint256(ORACLE_CONVERSION_W_BTC), 10 ** ORACLE_DECIMALS_BTC
+        )
+            + VAULT4_TOTAL_STAKE.mulDiv(OPERATOR5_SHARES_V4, VAULT4_TOTAL_SHARES).mulDiv(
+                uint256(ORACLE_CONVERSION_ST_ETH), 10 ** ORACLE_DECIMALS_ETH
+            );
+        assertEq(validators[4].power, expectedOperatorPower5);
+
+        // On Vault 4: delegator is operator specific and OP6 is assigned 1/3 shares
```
NindoK, line 630: Network restake* Not operator specific

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+        uint256 staker1Stake = 20_000 * 10 ** TOKEN_DECIMALS_USDC;
+        uint256 staker2Stake = 30_000 * 10 ** TOKEN_DECIMALS_USDC;
+
+        _depositToVault(vaultsData.v1.vault, staker1, staker1Stake, usdc, true);
+        _depositToVault(vaultsData.v1.vault, staker2, staker2Stake, usdc, true);
+
+        vm.warp(NETWORK_EPOCH_DURATION + 2);
+        Middleware.ValidatorData[] memory validators = middlewareReader.getValidatorSet(middleware.getCurrentEpoch());
+
+        uint256 expectedOperatorPower1 = (OPERATOR1_STAKE_V1_USDC + staker1Stake + staker2Stake).mulDiv(
```
NindoK, line 655: Let's add a comment that since it's operator specific v1 all the stake sums up on him

NindoK: Also another comment respecifying that operator1 is also on v2. Probably a small natspec on top would be good

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+        expectedOperatorPower1 -= operator1_withdraw_stake.mulDiv(10 ** 18, 10 ** TOKEN_DECIMALS_USDC); // Normalized to 18 decimals
+        validators = middlewareReader.getValidatorSet(middleware.getCurrentEpoch());
+        assertEq(validators[0].power, expectedOperatorPower1);
+    }
+
+    // ************************************************************************************************
+    // *                                       REWARDS DISTRIBUTION
+    // ************************************************************************************************
+
+    // We test each operator independently since rewards are accumulated in the stakers rewards contracts
```
NindoK, line 722: What about their rewards as operators? Why only as stakers?

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+            uint256 adminFeeStakerRewardsVault5 = expectedRewardsStakerVault5.mulDiv(ADMIN_FEE, MAX_PERCENTAGE);
+            expectedRewardsStakerVault5 -= adminFeeStakerRewardsVault5;
+
+            // Operator 7 in Vault 5
+            uint256 expectedRewards = expectedRewardsStakerVault5;
+            _checkClaimableRewards(stakerRewardsContractVault5, epoch, operator7, expectedRewards);
+        }
+    }
+
+    function testRewardsAreDistributedCorrectlyForOperator6() public {
```
NindoK, line 915: Change name since it reverts?

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+            assertEq(STAR.balanceOf(recipient), expectedAdminFee);
+
+            vm.stopPrank();
+        }
+
+        // Staker rewards contract should have distributed all the balance
+        assertEq(STAR.balanceOf(stakerRewardsContractVault2), 0);
+    }
+
+    function _loadRewardsRootAndProof(
```
NindoK, line 1143: Let's add all this at the top of the file right below setup, with the other internals function

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+        // Distribute rewards
+        uint48 eraIndex = 5;
+        uint256 amountToDistribute = 100 ether;
+        vm.warp(block.timestamp + 5 * NETWORK_EPOCH_DURATION);
+
+        uint48 epoch = _prepareRewardsDistribution(eraIndex, amountToDistribute);
+
+        uint256 initGas = gasleft();
+        uint256 expectedRewardsForStakers =
+            _claimAndCheckOperatorRewardsForOperator(amountToDistribute, eraIndex, OPERATOR8_KEY, operator8, 8, true);
```
NindoK, line 1398: Add a comment with current gas cost

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+    function _deployMiddlewareWithProxy(address operatorRewardsAddress, address stakerRewardsFactoryAddress) public {
+        IMiddleware.InitParams memory params = IMiddleware.InitParams({
+            network: tanssi,
+            operatorRegistry: address(operatorRegistry),
+            vaultRegistry: address(vaultFactory),
+            operatorNetworkOptIn: address(operatorNetworkOptInService),
+            owner: owner,
+            epochDuration: NETWORK_EPOCH_DURATION,
+            slashingWindow: SLASHING_WINDOW,
+            reader: address(0)
```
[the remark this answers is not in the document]
steven2308: To make it simpler, we already do this in many places. The method will create a reader if there is none.

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+        // Operator 1
+        rewardsOperator1 = expectedRewardsStakerVault2.mulDiv(OPERATOR1_STAKE_V2_WBTC, VAULT2_TOTAL_STAKE);
+        _checkClaimableRewards(stakerRewardsContractVault2, epoch, operator1, rewardsOperator1);
+
+        // Operator 2
+        rewardsOperator2 = expectedRewardsStakerVault2.mulDiv(OPERATOR2_STAKE_V2_WBTC, VAULT2_TOTAL_STAKE);
+        _checkClaimableRewards(stakerRewardsContractVault2, epoch, operator2, rewardsOperator2);
+
+        // Operator 3
+        rewardsOperator3 = expectedRewardsStakerVault2.mulDiv(OPERATOR3_STAKE_V2_WBTC, VAULT2_TOTAL_STAKE);
```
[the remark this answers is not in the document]
steven2308: No, we discussed this on Slack [here](https://moondance-labs.slack.com/archives/C063T9F091P/p1744637068077549). Limits and shares are taken into account to calculate power. To calculate rewards, only stake is used.

test/integration/Full.sol
steven2308: To make sure we get decimals right.  ETH and BTC have 18 decimals in the token, real usdc has 2.
Oracle decimals is independent, I went for 2 decimals for ETH and BTC, while real USDC oracle actually has 8.

test/integration/Full.sol
steven2308: Not sure I get your point, we pass address zero in almost every other test. 

I actually was wondering if we should pass vaults when the operator will only use that vault. Is that what you are suggesting?  
It's not the case for OP1 though, but for a few other operators it is.

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+        INetworkRestakeDelegator(vaultsData.v4.delegator).setOperatorNetworkShares(
+            tanssi.subnetwork(0), operator6, OPERATOR6_SHARES_V4
+        );
+        vm.stopPrank();
+    }
+
+    function _setLimitForNetworkAndOperators(
+        address _owner
+    ) public {
+        vm.startPrank(_owner);
```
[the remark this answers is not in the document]
steven2308: I'm already calling `setMaxNetworkLimit` and `setMaxNetworkLimit` for vaults 1 and 5 which are operator specific.
The `setOperatorNetworkLimit` method exists only in `FullRestakeDelegator`.

test/integration/Full.sol
steven2308: Why? Internal and private functions are supposed to go last.

src/contracts/middleware/Middleware.sol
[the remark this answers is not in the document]
NindoK: Yes but it's an issue that it's only with a local blockchain spin up by tests. An underflow with current timestamp on blockchain mainnet can't happen.

test/integration/Full.sol
NindoK: No wait I don't understand it xD Real usdc has 6 decimals right? And the oracle doesn't have these values or yes?

test/integration/Full.sol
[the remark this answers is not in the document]
NindoK:
```
Math.min(
                IVault(vault).activeStakeAt(timestamp, stakesHints.activeStakeHint),
                Math.min(
                    networkLimitAt(subnetwork, timestamp, stakesHints.networkLimitHint),
                    operatorNetworkLimitAt(subnetwork, operator, timestamp, stakesHints.operatorNetworkLimitHint)
                )
            ),
```

This is the code for the FullRestake. So basically the stake is the networkLimit/operatorNetworkLimit or activeStake of the whole vault. So no matter what if he has a limit of 2 it should be for 2, not for 3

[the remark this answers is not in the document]
NindoK: Oh yes, we use `activeSharesOfAt` for the vault. Makes sense

test/integration/Full.sol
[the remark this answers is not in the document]
NindoK: Oh yes. Remember thanks

test/integration/Full.sol
NindoK: The `registerOperator` have a third param which is the vault. If you look at the code you can only set this when the vault is `OperatorSpecific` or `NetworkOperatorSpecific`. So basically you don't need to register the vault (because it's not a shared vault), but actually register the operator with its vault.

test/integration/Full.sol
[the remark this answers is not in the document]
NindoK: Oh but then pelase change the interface. Don't call it from networkRestake

test/integration/Full.sol
NindoK: Then can we put it in the test title `testStakerRewards`. Otherwise Rewards is too generic

test/integration/Full.sol
```diff
@@ -0,0 +1,1474 @@
...
+
+        // Power changes even before the vault epoch ends, only network epoch needs to
+        vm.warp(block.timestamp + NETWORK_EPOCH_DURATION);
+        expectedOperatorPower1 -= operator1_withdraw_stake.mulDiv(10 ** 18, 10 ** TOKEN_DECIMALS_USDC); // Normalized to 18 decimals
+        validators = middlewareReader.getValidatorSet(middleware.getCurrentEpoch());
+        assertEq(validators[0].power, expectedOperatorPower1);
+    }
+
+    // ************************************************************************************************
+    // *                                       REWARDS DISTRIBUTION
```
NindoK, line 719: Even here if all of these tests are about stakers, just rename it

test/integration/Full.sol
NindoK: Because in all the tests we have used private functions that create a specific environment aka more like a "setup" right below setup. So when you read the code you first have the setup functions and then the tests.

test/integration/Full.sol
steven2308: Ok nice, will change.

src/contracts/middleware/Middleware.sol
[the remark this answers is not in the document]
steven2308: You are right, this could only be an actual issue if this was deployed in a newly created blockchain and on top of that we try to slash before slashing window. Will rollback.

test/unit/Deploy.t.sol
steven2308: This parameter is only used for deploying a vault, when delegator type is `operator specific`. So it can be zero as long as it is not the case.
Also, when delegator type is operator specific we must pass the `operator`, not the network nor the owner.

Same applies on other cases.

test/integration/Full.sol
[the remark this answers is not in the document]
steven2308: Deal

script/DeployVault.s.sol
steven2308: This was a nice one, had to adapt a few places so we could test this type of delegator. All good now!

test/integration/Full.sol
steven2308: We'll go with a 4th option similar to 2.
We will create a new role for `OperatorRewards` contract: `STAKER REWARDS SETTER`.  
The middleware will have this role and also an admin. When we need to use operator specific vaults, we will deploy the staker rewards independently and use the setter to link it to the vault.

NindoK (approved):

PR #74, opened 2025-04-11
Basic script to read specific storage location in a contract

NindoK:


utils/checkStorage.js
```diff
@@ -0,0 +1,180 @@
+require("dotenv").config();
```
steven2308, line 1: Broooo.. why not typescript? :cry:

steven2308 (commented):
Looks nice.. would look much better in typescript :grimacing:

utils/checkStorage.js
NindoK: Because I thought it's a simple script for a minor thing and making it typescript is an hassle that doesnt weigh in? 🤣

girazoki: it's a one time script, as long as it works and we know how to run it, I am fine with it.

steven2308 (approved):

PR #75, opened 2025-04-16
Added chainlink monitor script with slack integration

NindoK:
This PR adds script to monitor on chain event every time operators keys are sent to the gateway. 

It can:
- by passing a lookback period, it will fetch all the past events and check if there are any mismatch
- listen to all the new events emitted and compare the payloads being sent.

In a case of mismatch a message to slack is sent

Example of slack message:
Screenshot 2025-04-16 at 12 54 13

If the `sendCurrentOperatorsKeys` is called then another message will be shown
Screenshot 2025-04-17 at 01 09 06


utils/chainlink_monitor/chainlink_monitor_slack_alert.ts
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// Parse command line argument for lookback time
+function parseLookbackArgument(): number {
+  const args = process.argv.slice(2);
+  if (args.length === 0) {
+    return 0; // Default: no lookback
+  }
+
+  const lookbackSeconds = parseInt(args[1], 10);
```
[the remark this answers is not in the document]
NindoK: This is wrong since when calling the script we are passing as args the lookback in seconds. To do that (since we are using a `script` defined in `package.json` we need to use `-- 3600`. -- will be at args[0].

utils/chainlink_monitor/compare.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    const operatorsArray = [...sortedOperators];
+
+    operatorsArray.push("0xNewOperatorAddressHere");
+    operatorsArray.push("0xNewOperatorAddressHere");
+    operatorsArray.push("0xNewOperatorAddressHere2");
+
+    operatorsKeys.push("0x000000000000000000000000000000000000dEaD");
+    operatorsKeys.push("0xNewOperatorAddressHere");
+    operatorsKeys.push("0xNewOperatorAddressHere3");
```
steven2308, line 34: Why are you adding mock data on both arrays?

steven2308 (commented):
Pretty clean job, love it.
Once I understand why these mock data is added I'll approve.

steven2308 (approved):
Awesome!

steven2308 (approved):

steven2308 (approved):
This is my last approve today :laughing:

.github/workflows/demo-test.yml
```diff
@@ -1,12 +1,32 @@
...
 on:
+  push:
+    branches:
+      - staging
+      - main
   pull_request:
   workflow_dispatch:
 
 jobs:
+  detect-changes:
```
evgeny-s, line 12: I don't think we need an extra run here; use the [path filter](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore):
```
  pull_request:
    paths:
      - '**/*.sol'
```

evgeny-s:
**Note regarding CI fix for detecting changes for `*.sol` files:**
CI still expects this job to finish successfully.
So if we change not *.sol files, the job won’t trigger.
So we have 2 options:
- Keep the implementation with the paths filter, but we must trigger the jobs by changing `*.sol` files.
- Keep the previous implementation. It is a bit more flexible, but has more code.
So, if we are ok with option 1, we can change it later, because the less code the better.

evgeny-s (approved):

PR #76, opened 2025-04-16
Prevents registration of empty keys for operators.

steven2308:
Since empty keys are no longer an option, removes unreachable check on Reader and test.

Addreses https://github.com/moondance-labs/tanssi-sr-labs/issues/29

### DIFF
--- a/src/contracts/middleware/Middleware.sol
+++ b/src/contracts/middleware/Middleware.sol
@@ -39,6 +39,7 @@ import {ISlasher} from "@symbiotic/interfaces/slasher/ISlasher.sol";
 import {IVetoSlasher} from "@symbiotic/interfaces/slasher/IVetoSlasher.sol";
 import {Subnetwork} from "@symbiotic/contracts/libraries/Subnetwork.sol";
 import {Operators} from "@symbiotic-middleware/extensions/operators/Operators.sol";
+import {BaseOperators} from "@symbiotic-middleware/extensions/operators/BaseOperators.sol";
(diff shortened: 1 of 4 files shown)


src/contracts/middleware/Middleware.sol
```diff
@@ -439,6 +443,22 @@ contract Middleware is
...
 
+    /**
+     * @inheritdoc BaseOperators
+     */
+    function _beforeRegisterOperator(
+        address operator,
+        bytes memory key,
+        address
+    ) internal pure override notZeroAddress(operator) {
+        if (abi.decode(key, (bytes32)) == bytes32(0)) {
```
NindoK, line 454: If we do this, we should also prevent it when we update the key!

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -635,11 +635,6 @@ contract OBaseMiddlewareReader is MiddlewareStorage, EpochCapture, VaultManager,
                 ++i;
             }
             bytes32 key = abi.decode(getOperatorKeyAt(operator, epochStartTs), (bytes32));
-
-            if (key == bytes32(0)) {
```
NindoK, line 639: Otherwise we should keep this. (Im referring to the previous comment)

src/contracts/middleware/Middleware.sol
NindoK: Rechecked the code all good!

PR #81, opened 2025-04-23
Added initializeV2 to reinitialize operator rewards to have access control

NindoK:


src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -101,10 +102,35 @@ contract ODefaultOperatorRewards is
...
+     * @param operatorShare_ The share of the operator.
+     * @param admin The address of the admin.
+     */
+    function initializeV2(uint48 operatorShare_, address admin) public reinitializer(2) notZeroAddress(admin) {
+        if (operatorShare_ >= MAX_PERCENTAGE) {
+            revert ODefaultOperatorRewards__InvalidOperatorShare();
+        }
+
+        __UUPSUpgradeable_init();
+        __ReentrancyGuard_init();
```
steven2308, line 129: Do we actually need these 2? They already ran

NindoK: Yes because hypothetically we should only call v2 and not initialize from here on out if we would redeploy from scratch

steven2308: Seems like easy to forget. I'd rather have  `__AccessControl_init();` in the original initialize, and remove these 2 from here.  I would also remove the operator share setting, it is already set, this code does nothing, we just need to grant role to the admin. 

Also, it should be called by the owner only, otherwise anyone could add themselves as admin.

After we upgrade we need to remove this method anyway, once we remove it we will need to add the call to `__AccessControl_init()` on the original one.

steven2308 (changes requested):

src/contracts/rewarder/ODefaultOperatorRewards.sol
NindoK: You can't put that in the previous initialize. We can't reuse it. This is just the fix for this current scenario. The previous initialize should stay for backward compatibility and because it can be called only once (`initalizer` modifier). The v2 is just there for reinitializing this field on an already initialized contract

steven2308: Code as it is now won't work when we deploy on mainnet. 
`initialize` needs to have everything, including `__AccessControl_init`.
We won't use it now, because it already ran, we just need to run the `initializev2` method, and it only needs to have `__AccessControl_init`.

steven2308 (approved):

steven2308 (approved):

PR #82, opened 2025-04-23
Scripts and doc to upgrade Middleware, OperatorRewards and StakerRewards.

steven2308:


script/DeployTanssiEcosystem.s.sol
```diff
@@ -418,11 +418,51 @@ contract DeployTanssiEcosystem is Script {
         vm.stopBroadcast();
     }
 
+    function runUpgradeMiddleware(
```
NindoK, line 421: I have usually used broadcast in the name to make it more obvious.

utils/checkStorage.js
```diff
@@ -85,6 +87,12 @@ function getMappingStorageSlot(position) {
   );
 }
 
+function getNewMappingStorageSlot(position) {
+  return hexValue(
+    ethers.BigNumber.from(NEW_STORAGE_LOCATION).add(position)
```
NindoK, line 92: Instead of adding a new method, we could just pass the storage location directly into the previous function => 
`getMappingStorageSlot(position)` => `getMappingStorageSlot(location, position)`

NindoK (approved):

PR #88, opened 2025-04-30
[MD-1048] Run some basic static analyzers + [MD-972] Revisit integration tests to make sure we don't overprank

steven2308:
Breaking added because of this change on StakerRewards:

```
-        if (finalAmount == 0) {
+        if (finalAmount != amount) {
```

IMO this is a fix and not breaking but Benny suggested we added the tag.

### DIFF
--- a/test/integration/Middleware.t.sol
+++ b/test/integration/Middleware.t.sol
@@ -1098,12 +1098,11 @@ contract MiddlewareTest is Test {
         }
     }
 
-    function testGasFor100OperatorsIn3VaultsSorted() public {
+    function testGasFor10OperatorsIn3VaultsSorted() public {
         uint16 count = 100;
         _addOperatorsToNetwork(count);
 
         vm.warp(NETWORK_EPOCH_DURATION + 2);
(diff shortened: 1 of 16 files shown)


test/integration/Middleware.t.sol
```diff
@@ -1116,32 +1115,27 @@ contract MiddlewareTest is Test {
...
         OBaseMiddlewareReader(address(middleware)).getValidatorSet(currentEpoch);
         uint256 gasAfter = gasleft();
         uint256 gasNotSorted = gasBefore - gasAfter;
         console2.log("Total gas used for non sorted: ", gasNotSorted);
-
-        vm.stopPrank();
     }
 
-    function testGasFor250OperatorsIn3VaultsSorted() public {
+    function testGasFor25OperatorsIn3VaultsSorted() public {
```
NindoK, line 1134: Why this change? It's 250 Operators xD

steven2308: It is not, check the code. I updated the name of the test to match what is happening on the test.

PR #89, opened 2025-05-08
[MD-970] Testing upgrades in fork + [MD-1086] Fix fork tests in Sepolia

steven2308:
There is a line out of coverage as the result of fixing the fork test in sepolia, I'm working on that. The problem is that there is no Default Collateral Factory on Sepolia, so in that specific case we need to deploy one just like in local test chain, but this is only used on fork test, which are ignored on coverage.

For the upgrade test I'm using a bit of a hack. I'm mocking Benny's key to give admin permissions to a test account so it can run the upgrades on real deployed contracts in Sepolia. An alternative could be to directly modify storage to give admin rights, but I feel current approach is cleaner.


test/fork/Upgrades.t.sol
```diff
@@ -0,0 +1,119 @@
...
+import {Test} from "forge-std/Test.sol";
+
+import {Middleware} from "src/contracts/middleware/Middleware.sol";
+import {IOBaseMiddlewareReader} from "src/interfaces/middleware/IOBaseMiddlewareReader.sol";
+import {DeployRewards} from "script/DeployRewards.s.sol";
+import {DeployTanssiEcosystem} from "script/DeployTanssiEcosystem.s.sol";
+import {ODefaultOperatorRewards} from "src/contracts/rewarder/ODefaultOperatorRewards.sol";
+import {ODefaultStakerRewards} from "src/contracts/rewarder/ODefaultStakerRewards.sol";
+
+contract MiddlewareTest is Test {
```
NindoK, line 26: UpgradesTest

test/fork/Upgrades.t.sol
```diff
@@ -0,0 +1,119 @@
...
+        stakerRewards = ODefaultStakerRewards(stakerRewardsAddress);
+
+        deployRewards = new DeployRewards();
+        deployRewards.setIsTest(false);
+        deployTanssiEcosystem = new DeployTanssiEcosystem();
+        tanssi = IOBaseMiddlewareReader(address(middleware)).NETWORK();
+        admin = deployTanssiEcosystem.tanssi(); // Loaded from the env OWNER_PRIVATE_KEY
+
+        // Actual admin in the 3 contracts, we use its account to set the admin role to our admin which will run using broadcast
+        vm.startPrank(0x8f7b28C2A36E805F4024c1AE1e96a4B75E50A512);
```
NindoK, line 60: I'd just put owner and take it from the `chain_data.json` and use it here

NindoK (approved):

PR #91, opened 2025-05-15
[L-13] Missing executeSlash() Call After requestSlash() for Veto Slashing

steven2308:
Anyone can call the method.

### DIFF
--- a/src/contracts/middleware/Middleware.sol
+++ b/src/contracts/middleware/Middleware.sol
@@ -375,17 +375,17 @@ contract Middleware is
 
     /**
      * @dev Execute a slash with a given slash index using hints.
-     * @param slasher address of the slasher, must be an instance of IVetoSlasher
+     * @param vault The vault address, must have a veto slasher
      * @param slashIndex index of the slash request
      * @param hints hints for checkpoints' indexes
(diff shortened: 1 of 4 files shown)


src/contracts/middleware/Middleware.sol
```diff
@@ -373,6 +373,21 @@ contract Middleware is
...
+     * @param slashIndex index of the slash request
+     * @param hints hints for checkpoints' indexes
+     * @return slashedAmount virtual amount of the collateral slashed
+     */
+    function executeSlash(
+        address slasher,
+        uint256 slashIndex,
+        bytes calldata hints
+    ) external returns (uint256 slashedAmount) {
+        return IVetoSlasher(slasher).executeSlash(slashIndex, hints);
```
NindoK, line 388: You can just call `_executeSlash(vault, slashIndex, hints);` which come from the middleware sdk

NindoK (changes requested):
I'd rather use the already implemented version

src/contracts/middleware/Middleware.sol
steven2308: Nice catch. Changed!

NindoK (approved):

PR #92, opened 2025-05-15
Added rate limiting for sending operator keys to substrate

NindoK:
https://github.com/PashovAuditGroup/Tanssi_April25_MERGED/issues/83


src/contracts/middleware/MiddlewareStorage.sol
```diff
@@ -32,6 +33,7 @@ abstract contract MiddlewareStorage {
     uint8 public constant DEFAULT_DECIMALS = 18;
     uint256 public constant VERSION = 1;
     uint256 public constant PARTS_PER_BILLION = 1_000_000_000;
+    uint256 public constant MAX_BLOCK_INTERVAL = 50; // 600 seconds ≈ 10 minutes
```
steven2308, line 36: This could have better name. I think it should be MIN for starters. 
How about `MIN_INTERVAL_TO_SEND_OPERATOR_KEYS`?

test/unit/Middleware.t.sol
```diff
@@ -1484,10 +1487,30 @@ contract MiddlewareTest is Test {
 
         vm.store(address(middleware), slot, bytes32(0));
 
+        vm.roll(80);
+
         vm.expectRevert(IMiddleware.Middleware__GatewayNotSet.selector);
         middleware.sendCurrentOperatorsKeys();
     }
 
+    function testSendCurrentOperatorKeysButIsLimitedByMaxIntervalBlock() public {
```
steven2308, line 1496: Would also change name from Max to Min. Because that is the minimum number of blocks you need to wait.

steven2308 (approved):
Minor suggestion to change name. All good besides that.

PR #106, opened 2025-05-23
[MD-1153] Clean deploy script for production

steven2308:


script/Deploy.sol
```diff
@@ -0,0 +1,170 @@
...
+            slashingWindow: SLASHING_WINDOW,
+            reader: reader
+        });
+
+        Middleware middlewareImpl = new Middleware(operatorRewardsAddress, stakerRewardsFactoryAddress);
+        middleware = Middleware(address(new MiddlewareProxy(address(middlewareImpl), "")));
+        middleware.initialize(params);
+
+        // All these needs to be called by the admin, we could make ourselves admins initially to do so:
+        middleware.setForwarder(entities.forwarder);
```
NindoK, line 103: These 2 won't happen. So we can comment it out just to remember it

script/Deploy.sol
```diff
@@ -0,0 +1,170 @@
...
+        middleware.setGateway(entities.gateway);
+        middleware.grantRole(middleware.DEFAULT_ADMIN_ROLE(), entities.admin);
```
NindoK, line 105: We will revoke the role to the current initialAdmin once we know everything is good. Let's put a comment about it to avoid confusion

script/Deploy.sol
steven2308: Also gateway? I thought we would deploy gateway first. 
If not, why?

NindoK: Because gateway will be deployed once Tanssi will go live. We will deploy middleware only for registering vaults and operators

script/DeployProduction.s.sol
```diff
@@ -0,0 +1,191 @@
...
+import {DeployRewards} from "./DeployRewards.s.sol";
+
+// Sessions: 1h
+// Eras: 6h
+// Middleware epochs: 1d
+// Slashing window: 2d
+// Vault epoch: 7d
+
+contract DeployProduction is Script {
+    uint48 public constant VAULT_EPOCH_DURATION = 7 days;
```
NindoK, line 38: You can take out not used

PR #115, opened 2025-06-05
Adds proposal to keep track of contract addresses, changelog and deployment logs per network.

steven2308:
I'm using a standard changelog with semantic versioning. This would come together with tags on the commits for each release. For each network there is a log file with the version and the date of deployment.

I started with version 1.0.0 being the one where we fixed the most important audit findings. After that there was a migration removing release (1.0.1) and then another one adding more funcionality and a fix (1.1.0).

### DIFF
--- a/deployment-logs/stagelight.log
+++ b/deployment-logs/stagelight.log
@@ -1,3 +1,3 @@
-- v1.1.0 deployed on 2025-06-03
-- v1.0.1 deployed on 2025-05-26
-- v1.0.0 deployed on 2025-05-19
+- v1.1.0 deployed on 2025-06-03 by @steven2308
+- v1.0.1 deployed on 2025-05-26 by @steven2308
+- v1.0.0 deployed on 2025-05-19 by @NindoK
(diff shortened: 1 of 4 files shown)


NindoK (approved):

deployment-logs/stagelight.log
```diff
@@ -0,0 +1,3 @@
+- v1.1.0 deployed on 2025-06-03
```
evgeny-s, line 1: Can we also add a commit sha, that has been deployed, also maybe the github-ID of the guy who deployed it.

steven2308: I'll add commit and deployer 100%.
About the logs, I feel it can bee too much. WDYT @NindoK?

steven2308: Actually, I think the commit will fit better in the Changelog. Since it's there where we define versions.

evgeny-s (approved):

deployment-logs/stagelight.log
evgeny-s: About logs, guys. Depend on how much it is, but I still think it might be really helpful.

PR #124, opened 2025-06-16
Initial CL Optimization version

NindoK:
Optimization managed to reduced gas cost for CL from `11M` to `3_265_029` for 100 operators in 3 vaults each.


src/contracts/middleware/Middleware.sol
```diff
@@ -14,6 +14,7 @@
 // along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
 pragma solidity 0.8.25;
 
+import {console2} from "forge-std/console2.sol";
```
steven2308, line 17: We can take this off now :smile:

src/contracts/middleware/Middleware.sol
```diff
@@ -306,13 +307,55 @@ contract Middleware is
...
+
+        address[] memory activeOperators = _activeOperators();
+        uint256 activeOperatorsLength = activeOperators.length;
+        if (activeOperatorsLength == 0) {
+            // No active operators, no upkeep needed
+            return (false, hex"");
+        }
+
+        uint256 cachedCurrentEpochValidatorsLength = cache.epochToValidatorsData[epoch].length;
+        uint256 diffLength = activeOperatorsLength - cachedCurrentEpochValidatorsLength;
```
steven2308, line 323: I'd call this something more informative. Like `pendingOperatorsToCache`

src/contracts/middleware/Middleware.sol
```diff
@@ -306,13 +307,55 @@ contract Middleware is
...
+            uint256 maxNumOperatorsToCheck = Math.min(diffLength, MAX_OPERATORS_TO_PROCESS);
+
+            ValidatorData[] memory validatorsData = new ValidatorData[](maxNumOperatorsToCheck);
+
+            // Populate validatorsData with the new operators' keys and their powers
+            // It gets encoded to be used in performUpkeep
+            for (
+                uint256 i = cachedCurrentEpochValidatorsLength;
+                i < cachedCurrentEpochValidatorsLength + maxNumOperatorsToCheck && i < activeOperatorsLength;
+                i++
```
steven2308, line 336: Not sure if this is efficient on 0.8.25, or if it's still better to do the `unchecked i++`

src/contracts/middleware/Middleware.sol
```diff
@@ -306,13 +307,55 @@ contract Middleware is
...
+            ) {
+                address operator = activeOperators[i];
+                bytes32 operatorKey = abi.decode(operatorKey(operator), (bytes32));
+                uint256 operatorPower = _getOperatorPowerAt(currentEpochStartTs, operator);
+                validatorsData[i - cachedCurrentEpochValidatorsLength] =
```
steven2308, line 341: Isn't it a + instead?
```suggestion
                validatorsData[i + cachedCurrentEpochValidatorsLength] =
```

src/contracts/middleware/Middleware.sol
```diff
@@ -327,15 +370,33 @@ contract Middleware is
         if (gateway == address(0)) {
             revert Middleware__GatewayNotSet();
         }
+        uint48 epoch = getCurrentEpoch();
+        StorageMiddlewareCache storage cache = _getMiddlewareStorageCache();
+
+        address[] memory activeOperators = _activeOperators();
+        uint256 activeOperatorsLength = activeOperators.length;
+        uint256 cachedCurrentEpochValidatorsLength = cache.epochToValidatorsData[epoch].length;
+        uint256 diffLength = activeOperatorsLength - cachedCurrentEpochValidatorsLength;
```
steven2308, line 379: Same suggestion about variable name

src/contracts/middleware/Middleware.sol
```diff
@@ -327,15 +370,33 @@ contract Middleware is
...
+
+        if (diffLength > 0) {
```
steven2308, line 381: I'm worried that this could be called twice with the same batch of operators. 
I know it costs more gas, but I'd rather not reuse the same cache to store the initial ones and then the sorted ones.  Or maybe to save a bit of gas, instead of duplicating we could just store the sorted indexes?
I can explain more in a call.

steven2308 (changes requested):

src/contracts/middleware/Middleware.sol
steven2308: Found this on CHANGELOG for 0.8.22. So unchecked is probably not needed anymore. Will try to test the gas usage today.
> Code Generator: Remove redundant overflow checks of certain for loops when the counter variable cannot overflow.

src/contracts/middleware/Middleware.sol
```diff
@@ -306,13 +307,55 @@ contract Middleware is
...
+        if (activeOperatorsLength == 0) {
+            // No active operators, no upkeep needed
+            return (false, hex"");
+        }
+
+        uint256 cachedCurrentEpochValidatorsLength = cache.epochToValidatorsData[epoch].length;
+        uint256 diffLength = activeOperatorsLength - cachedCurrentEpochValidatorsLength;
+
+        // Check if cache is still not filled with the current epoch validators
+        if (diffLength > 0) {
```
evgeny-s, line 326: Can it be the situation when `activeOperators` < cached amount? Let's say 2 operators have been removed and 1 added, so we need to update the cache, but diffLength will be < 0.

src/contracts/middleware/Middleware.sol
evgeny-s: Same here as above

src/contracts/middleware/Middleware.sol
NindoK: Imho wouldn't be possible since: at the start of epoch you would get activeOperators and from that point on operators number can't change so would be impossible to get cached > active operators. Nice invariant to check nonetheless

src/contracts/middleware/Middleware.sol
NindoK: Nope, because this should go from 0 to MAX_OPERATORS_TO_ADD since the `validatorsData` array contains at most the min between the max and the current left.

src/contracts/middleware/Middleware.sol
NindoK: Maybe we can just check that if the power has been already set for that operator we can just revert the whole execution. This will trigger an alert on our script.

src/contracts/middleware/Middleware.sol
steven2308: Let's leave this one unresolved so we don't forget to take it out when it's ready

src/contracts/middleware/Middleware.sol
```diff
@@ -306,13 +307,55 @@ contract Middleware is
...
+                    ValidatorData({key: operatorKey, power: operatorPower});
+            }
+            // encode values to be used in performUpkeep
+            return (true, abi.encode(validatorsData));
+        }
 
         //Should be at least once per epoch
         upkeepNeeded = (Time.timestamp() - $.lastTimestamp) > $.interval;
+        if (upkeepNeeded) {
+            // This will use the cached values, resulting in just a simple sorting operation. We can know a priori how much it cost since it's just an address with a uint256 power. Worst case we can split this too.
```
steven2308, line 351: I don't understand this.  Why is it just one?

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -260,28 +261,35 @@ contract ODefaultOperatorRewards is
...
         uint256 totalPower;
-        for (uint256 i; i < totalVaults;) {
-            // TODO: Optimize using the cache
-            vaultPowers[i] = reader.getOperatorPowerAt(epochStartTs, operator, operatorVaults[i], subnetwork);
-            totalPower += vaultPowers[i];
-            unchecked {
-                ++i;
+        uint256[] memory vaultPowers;
+        {
+            bytes32 operatorKey = abi.decode(Middleware(middlewareAddress).operatorKey(operator), (bytes32));
```
steven2308, line 274: Here you are using the operator key which comes from substrate AFAICS. But when you cache it, you just encode the evm address as bytes32.  We need to use the same in both places, it looks like you already have the evm address here and there so maybe we can have the key of the cache be the address directly.

NindoK: We don't want the address otherwise we go back to the same issue that was found by Pashov. We only use operatorKey as such we never have a problem if someone changes it and it's compatible in history

src/contracts/middleware/MiddlewareStorage.sol
```diff
@@ -160,8 +161,13 @@ abstract contract MiddlewareStorage {
...
-    function getOperatorToPower(uint48 epoch, bytes32 operatorKey) public view returns (uint256 power) {
+    function getOperatorToPower(
+        uint48 epoch,
+        bytes32 operatorKey
+    ) public view returns (uint256 power, uint256[] memory powerPerVault) {
         StorageMiddlewareCache storage $ = _getMiddlewareStorageCache();
-        power = $.operatorKeyToPower[epoch][operatorKey];
+        IMiddleware.ValidatorDataCache memory operatorPowerData = $.operatorKeyToPower[epoch][operatorKey];
+        power = operatorPowerData.power;
+        powerPerVault = operatorPowerData.powerPerVault;
```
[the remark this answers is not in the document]
NindoK: Don't understand this. Could you clarify?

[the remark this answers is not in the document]
steven2308: I mean, what if are trying to read the cache but it wasn't cached already?

src/contracts/middleware/MiddlewareStorage.sol
```diff
@@ -160,8 +161,13 @@ abstract contract MiddlewareStorage {
      * @param operatorKey The operator key
      * @return power The power of the operator
      */
-    function getOperatorToPower(uint48 epoch, bytes32 operatorKey) public view returns (uint256 power) {
+    function getOperatorToPower(
+        uint48 epoch,
+        bytes32 operatorKey
+    ) public view returns (uint256 power, uint256[] memory powerPerVault) {
```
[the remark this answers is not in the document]
steven2308: But you never need both, so you are not really optimizing gas.  Maybe size? But may not even be the case, I've had times where it takes less size to have 2 getter/setter methods instead of one with 2 or multiple.

src/contracts/middleware/Middleware.sol
```diff
@@ -327,15 +370,33 @@ contract Middleware is
         if (gateway == address(0)) {
             revert Middleware__GatewayNotSet();
         }
+        uint48 epoch = getCurrentEpoch();
```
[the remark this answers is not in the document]
steven2308: I just had the same question haha.
But even if it happens, we would start caching the new epoch and ignore the previous one.

src/contracts/middleware/MiddlewareStorage.sol
```diff
@@ -26,14 +28,25 @@ abstract contract MiddlewareStorage {
         uint256 lastExecutionBlock;
     }
 
+    struct StorageMiddlewareCache {
+        mapping(uint48 epoch => uint256 cacheIndex) epochToCacheIndex;
+        mapping(uint48 epoch => mapping(bytes32 operatorKey => IMiddleware.ValidatorData operatorPower))
+            operatorKeyToPower;
+    }
```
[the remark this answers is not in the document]
NindoK: It made sense before. Here no! Thanks :D

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -692,4 +677,62 @@ contract OBaseMiddlewareReader is
...
+     * @return performData bytes of the sorted (by power) operators' keys and the epoch that will be used by the keeper when calling performUpkeep, if upkeep is needed.
+     */
+    function auxiliaryCheckUpkeep() external view returns (bool upkeepNeeded, bytes memory performData) {
+        uint48 epoch = getCurrentEpoch();
+        uint48 currentEpochStartTs = getEpochStart(epoch);
+
+        StorageMiddleware storage $ = _getMiddlewareStorage();
+        StorageMiddlewareCache storage cache = _getMiddlewareStorageCache();
+
+        address[] memory activeOperators = _activeOperators();
```
steven2308, line 694: This shadows the getter. Let's name it `activeOperators_`

src/contracts/middleware/Middleware.sol
```diff
@@ -322,20 +316,54 @@ contract Middleware is
...
-            $.lastTimestamp = currentTimestamp;
+        uint48 epoch = getCurrentEpoch();
+        StorageMiddlewareCache storage cache = _getMiddlewareStorageCache();
+
+        address[] memory activeOperators = _activeOperators();
+        uint256 activeOperatorsLength = activeOperators.length;
+        uint256 cacheIndex = cache.epochToCacheIndex[epoch];
+        uint256 pendingOperatorsToCache = activeOperatorsLength - cacheIndex;
+
+        if (pendingOperatorsToCache > 0) {
```
steven2308, line 337: I would like us to have a test for the following case:
1. Make sure all the operators are already cached.
2. Repeat the last call with ValidatorsData.

Since all operators are cached already, it will go the second branch of the if, trying to decode as an array of bytes32. I want to make sure that fails and we decode it as garbage by mistake.

steven2308 (approved):

PR #126, opened 2025-06-20
Feat/mainnet fork tests

steven2308:


test/fork/mainnet/Full.t.sol
```diff
@@ -1269,209 +1388,473 @@ contract FullTest is Test {
...
+        }
+
+        uint256 expectedStakerRewards =
+            _claimAndCheckRewardsForOperator(eraIndex, operator, proofAndPoints, totalPoints);
+
+        for (uint256 i; i < totalVaults; i++) {
+            uint256 expectedStakerRewardsForVault = expectedStakerRewards.mulDiv(operator.powers[i], totalPower);
+            uint256 newStakerRewardsBalancesBefore = rewardsToken.balanceOf(vaultToStakerRewards[operator.vaults[i]]);
+            uint256 actualRewards = newStakerRewardsBalancesBefore - stakerRewardsBalancesBefore[i];
+            // TODO: For some reason one of them has a difference of 6, so we allow a difference of up to 10. It's a millionth of a cent but could be worth checking since it's only for one.
```
NindoK, line 1739: Which token and in which case? Might be important

test/fork/mainnet/Full.t.sol
```diff
@@ -1269,209 +1388,473 @@ contract FullTest is Test {
...
+        vm.stopPrank();
+    }
+
+    function _testSlashingAndExecutingSlashForOperator(
+        HelperConfig.OperatorData memory operator
+    ) public {
+        uint48 initialEpoch = middleware.getCurrentEpoch();
+        uint48 epochStartTs = reader.getEpochStart(initialEpoch);
+
+        // We need to track by stake and not by power since power uses live oracle which cannot be mocked
```
NindoK, line 1836: We can use live data tho right? We can make direct call to live oracles

test/fork/mainnet/Full.t.sol
steven2308: The token is TANSSI, which has 12 decimals. The deviation depends on whatever the current stake values are, I just checked and there are differences in 7 cases, ranging from 1 to 4.  That is 4x10e-12.  
I think we are safe without the TODO. 
WDYT?

NindoK: It's fine by me. I just want to understand why these errors. Differences of 1 wei is fine, because it can be rounding error. But greater than 1 it's strange

test/fork/mainnet/Full.t.sol
```diff
@@ -1269,209 +1388,473 @@ contract FullTest is Test {
...
+    }
+
+    function testSlashingAndExecutingSlashForOperator11Opslayer() public {
+        _testSlashingAndExecutingSlashForOperator(operators.operator11Opslayer);
+    }
+
+    function testOperatorRewardsDistributionCase1() public {
+        (uint48 eraIndex, uint32 totalPoints) = _prepareRewardsDistributionForCase(1);
+
+        assertNotEq(totalPoints, 0);
```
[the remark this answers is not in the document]
steven2308: Done

test/fork/mainnet/Full.t.sol
```diff
@@ -1269,209 +1388,473 @@ contract FullTest is Test {
...
+        );
+        _claimAndCheckRewardsForOperator(
+            eraIndex, operators.operator3Nodeinfra, proofAndPointsByOperator.operator3Nodeinfra, totalPoints
+        );
+        _claimAndCheckRewardsForOperator(
+            eraIndex, operators.operator4Blockscape, proofAndPointsByOperator.operator4Blockscape, totalPoints
+        );
+    }
+
+    function testOperatorRewardsDistributionCase2() public {
```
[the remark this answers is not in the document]
steven2308: Added comments, they are pretty simple, no need to maintain since it's just which operators are tested.

NindoK (approved):

PR #127, opened 2025-06-30
[MD-1185] - Create test that proves if operator is unregistered no vaults are shown in the past

NindoK:


test/integration/Middleware.t.sol
```diff
@@ -554,6 +554,41 @@ contract MiddlewareTest is Test {
...
+        Middleware.ValidatorData[] memory validators =
+            OBaseMiddlewareReader(address(middleware)).getValidatorSet(currentEpoch);
+
+        operatorVaults = OBaseMiddlewareReader(address(middleware)).activeVaults(operator2);
+
+        uint256 operatorPower =
+            OBaseMiddlewareReader(address(middleware)).getOperatorPowerAt(previousEpochStartTs, operator2);
+        address[] memory previousActiveOperators =
+            OBaseMiddlewareReader(address(middleware)).activeOperatorsAt(previousEpochStartTs);
+        assertGt(operatorPower, 0);
```
steven2308, line 581: So power and vaults are kept, right?

steven2308 (approved):
So if we unregister operators we will not be able to get them anymore from:
- getOperatorsByEpoch 
- getValidatorSet
- getOperatorVaultPairs
- getTotalStake

This sucks but let's think when is this a real problem. I think we just need these methods to send operators data to substrate, but we will always use the current epoch.

We also use the reader when distributing rewards for operator, to know the participation of each vault. But there we use getOperatorPowerAt. Since power is not erased we should be good too.

Is there any other place we use it?

NindoK:
> So if we unregister operators we will not be able to get them anymore from:
> 
> * getOperatorsByEpoch
> * getValidatorSet
> * getOperatorVaultPairs
> * getTotalStake
> 
> This sucks but let's think when is this a real problem. I think we just need these methods to send operators data to substrate, but we will always use the current epoch.

Yes, so there is not a proper issue on that side.

> 
> We also use the reader when distributing rewards for operator, to know the participation of each vault. But there we use getOperatorPowerAt. Since power is not erased we should be good too.
> 

100%, fortunately we don't have this issue either

> Is there any other place we use it?

Doesn't look like it. Just double checked again

PR #132, opened 2025-07-04
[MD-1212] - Optimize staker rewards to set the vault, instead of passing it through constructor

NindoK:
This introduces a new way to think about `ODefaultStakerRewardsFactory`.

It will require some new initialization to set `i_operatorRewards` and `i_stakerRewardsFactory`


src/interfaces/rewarder/IODefaultStakerRewardsFactory.sol
```diff
@@ -18,9 +18,43 @@ import {IODefaultStakerRewards} from "./IODefaultStakerRewards.sol";
...
     error ODefaultStakerRewardsFactory__InvalidAddress();
+    error ODefaultStakerRewardsFactory__InvalidImplementation();
+    error ODefaultStakerRewardsFactory__InvalidVersion();
     error ODefaultStakerRewardsFactory__NotVault();
 
+    /**
+     * @notice Emitted when a new implementation is whitelisted.
+     * @param implementation address of the new implementation
+     */
+    event Whitelist(address indexed implementation);
```
steven2308, line 31: I would rename all mentions of whitelist to newImplementation.

src/contracts/rewarder/ODefaultStakerRewardsFactory.sol
```diff
@@ -41,6 +53,38 @@ contract ODefaultStakerRewardsFactory is Registry, IODefaultStakerRewardsFactory
...
+        return _whitelistedImplementations.at(version - 1);
+    }
+
+    /**
+     * @inheritdoc IODefaultStakerRewardsFactory
+     */
+    function whitelist(
+        address implementation_
+    ) external onlyOwner {
+        if (!_whitelistedImplementations.add(implementation_)) {
```
steven2308, line 81: Feels too much IMO, I would just keep the latest one and emit an event.

test/unit/Middleware.t.sol
```diff
@@ -212,12 +213,19 @@ contract MiddlewareTest is Test {
             owner: owner,
             epochDuration: NETWORK_EPOCH_DURATION,
             slashingWindow: SLASHING_WINDOW,
-            reader: readHelper
+            reader: readHelper,
+            operatorRewards: operatorRewardsAddress,
+            stakerRewardsFactory: stakerRewardsFactoryAddress
         });
         middleware.initialize(params);
+        // This should be taken out later on
```
steven2308, line 221:
```suggestion
        // TODO: This should be taken out later on
```
So we can find it

src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -339,6 +340,21 @@ contract ODefaultStakerRewards is
         emit SetAdminFee(adminFee_);
     }
 
+    /**
+     * @dev This function should be called only once and then should be upgraded to a new implementation without this function.
```
steven2308, line 344:
```suggestion
     * @dev TODO: This function should be called only once and then should be upgraded to a new implementation without this function.
```
So we remember. Should also be included in migration, together with calling initializev2 on Middleware.

steven2308 (changes requested):
Huge improvement to have a single implementation for staker rewards.
Also removing constructor params from middleware is an epic unexpected win.

steven2308 (approved):

test/unit/Middleware.t.sol
```diff
@@ -215,9 +216,14 @@ contract MiddlewareTest is Test {
             reader: readHelper
         });
         middleware.initialize(params);
+        //TODO: This should be taken out later on
```
steven2308, line 219: According to our talk yesterday, we can remove this TODO

PR #135, opened 2025-07-08
Feat/introduce obbligatory migration label

NindoK:
This PR introduces a new step in the CI job to check for the presence of one of the two labels: `need-migration` or `no-migration`. 

If `need-migration` is present then changes in MIGRATION.md must be made.

It enforces that each PR needs to be thought through, adding another steps in maintaining full clearance on what needs to be migrated and how


steven2308 (approved):
LGTM!
I'd like to have @evgeny-s approve as well.

evgeny-s (approved):
LGTM

PR #141, opened 2025-07-14
[MD-1183] Optimizes gas on checkUpkeep and prevents operators on too many vaults.

steven2308:
We cannot control opt ins from operators to vaults, but we can control how many shared vaults we let in, and how many operator vaults also.  I added a check on hooks we were already using for the registration of both types of vaults. 
The check was added in the reader to save space on Middleware.

Even with the checks on place, it can happen that an operator registers to some operator specific vaults and then into many shared vaults. This case cannot be prevented on registration so we also add an extra safe net when getting operators for upkeep. If at the time an operator can be active in more than the limit of vaults, we set its power to zero, so it is in the last position and substrate site hopefully ignores it.

This makes me think again that we should have a power threshold to send operators to substrate. We have talked in the pass about it and the consensus has been to not do anything, but this extra case might turn the tides.

### DIFF
--- a/src/contracts/middleware/Middleware.sol
+++ b/src/contracts/middleware/Middleware.sol
@@ -85,17 +85,9 @@ contract Middleware is
 
     /*
      * @notice Constructor for the middleware
-     * @param operatorRewards The operator rewards address
-     * @param stakerRewardsFactory The staker rewards factory address
      */
-    constructor(
-        address operatorRewards,
-        address stakerRewardsFactory
-    ) notZeroAddress(operatorRewards) notZeroAddress(stakerRewardsFactory) {
+    constructor() {
         _disableInitializers();
-
-        i_operatorRewards = operatorRewards;
-        i_stakerRewardsFactory = stakerRewardsFactory;
     }
 
     /*
@@ -138,6 +130,19 @@ contract Middleware is
         _setSelectorRole(this.performUpkeep.selector, FORWARDER_ROLE);
     }
 
+    /*
+     * @notice Reinitialize the middleware with only operator rewards and staker rewards factory addresses
+     * @param operatorRewards The operator rewards address
+     * @param stakerRewardsFactory The staker rewards factory address
+     */
+    function reinitializeRewards(
+        address operatorRewards,
+        address stakerRewardsFactory
+    ) external reinitializer(2) notZeroAddress(operatorRewards) notZeroAddress(stakerRewardsFactory) {
+        i_operatorRewards = operatorRewards;
+        i_stakerRewardsFactory = stakerRewardsFactory;
+    }
+
     function _validateInitParams(
         InitParams memory params
     )
@@ -160,6 +165,10 @@ contract Middleware is
     }
 
     function stakeToPower(address vault, uint256 stake) public view override returns (uint256 power) {
+        if (stake == 0) {
+            return 0;
+        }
+
         address collateral = vaultToCollateral(vault);
         address oracle = collateralToOracle(collateral);
 
@@ -531,7 +540,7 @@ contract Middleware is
         bytes memory key,
         address
     ) internal pure override notZeroAddress(operator) {
-        if (abi.decode(key, (bytes32)) == bytes32(0)) {
+        if (key.length != 32 || abi.decode(key, (bytes32)) == bytes32(0)) {
             revert Middleware__InvalidKey();
         }
     }
--- a/src/contracts/middleware/Middleware.sol
+++ b/src/contracts/middleware/Middleware.sol
@@ -510,7 +510,7 @@ contract Middleware is
     /**
      * @inheritdoc BaseOperators
      */
-    function _beforeRegisterOperatorVault(address operator, address vault) internal override {
+    function _beforeRegisterOperatorVault(address, /* operator */ address vault) internal override {
         _setVaultToCollateral(vault);
     }
 
--- a/src/contracts/middleware/Middleware.sol
+++ b/src/contracts/middleware/Middleware.sol
@@ -18,13 +18,11 @@ pragma solidity 0.8.25;
 //                                      CHAINLINK
 //**************************************************************************************************
 import {AutomationCompatibleInterface} from "@chainlink/automation/interfaces/AutomationCompatibleInterface.sol";
-import {AggregatorV3Interface} from "@chainlink/shared/interfaces/AggregatorV2V3Interface.sol";
 
 //**************************************************************************************************
 //                                      OPENZEPPELIN
 //**************************************************************************************************
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import {IERC20Metadata} from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
 import {Time} from "@openzeppelin/contracts/utils/types/Time.sol";
 import {UUPSUpgradeable} from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
@@ -165,24 +163,7 @@ contract Middleware is
     }
 
     function stakeToPower(address vault, uint256 stake) public view override returns (uint256 power) {
-        if (stake == 0) {
-            return 0;
-        }
-
-        address collateral = vaultToCollateral(vault);
-        address oracle = collateralToOracle(collateral);
-
-        if (oracle == address(0)) {
-            revert Middleware__NotSupportedCollateral(collateral);
-        }
-        (, int256 price,,,) = AggregatorV3Interface(oracle).latestRoundData();
-        uint8 priceDecimals = AggregatorV3Interface(oracle).decimals();
-        power = stake.mulDiv(uint256(price), 10 ** priceDecimals);
-        // Normalize power to 18 decimals
-        uint8 collateralDecimals = IERC20Metadata(collateral).decimals();
-        if (collateralDecimals != DEFAULT_DECIMALS) {
-            power = power.mulDiv(10 ** DEFAULT_DECIMALS, 10 ** collateralDecimals);
-        }
+        return IOBaseMiddlewareReader(address(this)).getPowerInUSD(vault, stake);
     }
 
     /**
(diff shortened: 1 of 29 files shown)


src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -735,4 +724,82 @@ contract OBaseMiddlewareReader is
...
+        address operator
+    ) private view returns (uint256 power) {
+        VaultManagerStorage storage $ = _getVaultManagerStorage();
+        address[] memory operatorVaults = $._operatorVaults[operator].getActive(timestamp);
+
+        uint256 totalSharedVaults = sharedVaults.length;
+        uint256 totalOperatorVaults = operatorVaults.length;
+
+        // This check might seem innecesary since we check on vault registration, however if we register first operator vaults and then shared ones, the limit might be reached for an operator without triggering the revert on registration.
+        if (totalSharedVaults + totalOperatorVaults <= MAX_ACTIVE_VAULTS) {
```
NindoK, line 772: How this is needed if we have the same condition on the registration process?

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -735,4 +724,82 @@ contract OBaseMiddlewareReader is
...
+            address[] memory vaults = new address[](totalSharedVaults + totalOperatorVaults);
+
+            for (uint256 i; i < totalSharedVaults; ++i) {
+                vaults[i] = sharedVaults[i];
+            }
```
NindoK, line 777: I feel like we can do something more here to avoid this. Feels kind of a waste :/ 

What if we create a big array and then we use:

```

        assembly ("memory-safe") {
            mstore(array, len)
        }
```
        
So we make sure that there are always enough and then we shrink it down. Might be cheaper. Can you try it?

src/contracts/middleware/Middleware.sol
```diff
@@ -505,6 +505,8 @@ contract Middleware is
         address sharedVault,
         IODefaultStakerRewards.InitParams memory stakerRewardsParams
     ) internal override {
+        IOBaseMiddlewareReader(address(this)).checkTotalActiveVaults(address(0));
```
NindoK, line 508: Uhu? What's the point here?

src/contracts/middleware/OBaseMiddlewareReader.sol
steven2308: Literally, read the comment 😂

src/contracts/middleware/OBaseMiddlewareReader.sol
steven2308: It was my initial idea. But I'd need to shorten it but maybe also lengthen it for the next operator which could break.

src/contracts/middleware/Middleware.sol
steven2308: To prevent vault registration if we hit the limit

src/contracts/middleware/OBaseMiddlewareReader.sol
NindoK: Can't be more than 100 right? So we already have an upper limit. You set to 100 and then you shrink it down

steven2308: Let's discuss tomorrow, not sure if we are thinking about the same, but I'm afraid the approach I have in mind can be problematic.

src/contracts/middleware/Middleware.sol
steven2308: To clarify, `checkTotalActiveVaults` always checks for shared vaults + operator vaults if an operator is passed. 
Since here it's the registration of a shared vault I don't do any check for operators.

src/contracts/middleware/OBaseMiddlewareReader.sol
steven2308: The second added test should help clarify this case.

src/contracts/middleware/Middleware.sol
```diff
@@ -525,7 +527,8 @@ contract Middleware is
     /**
      * @inheritdoc BaseOperators
      */
-    function _beforeRegisterOperatorVault(address, /* operator */ address vault) internal override {
+    function _beforeRegisterOperatorVault(address operator, address vault) internal override {
+        IOBaseMiddlewareReader(address(this)).checkTotalActiveVaults(operator);
```
NindoK, line 531: What if we don't control it, but we just don't account it when calculate the power? So if someone has too many vaults we just skip them?

Just a thought to think about, mainly because I don't want to lower the optimizations run

steven2308: I can buy it for operator vaults. But I would def check for shared vaults because once we go over the limit we will ignore all of the operators when reading on checkUpkeep.

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -58,16 +64,37 @@ contract OBaseMiddlewareReader is
     using Subnetwork for bytes32;
     using Math for uint256;
 
-    // ** OLD BASE MIDDLEWARE READER LOGIC **
```
NindoK, line 61: I'd keep this for where is needed. So we know what's in the basic and what's what we added

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -735,4 +749,51 @@ contract OBaseMiddlewareReader is
...
+        uint48 timestamp,
+        address[] memory activeOperators_
+    ) private view returns (IMiddleware.ValidatorData[] memory validatorsData) {
+        // Populate validatorsData with the new operators' keys and their powers
+        // It gets encoded to be used in performUpkeep
+        validatorsData = new IMiddleware.ValidatorData[](maxNumOperatorsToCheck);
+
+        VaultManagerStorage storage $ = _getVaultManagerStorage();
+        address[] memory sharedVaults = $._sharedVaults.getActive(timestamp);
+        uint160[] memory subnetworks = _activeSubnetworksAt(timestamp);
```
NindoK, line 765: Since we have a single subnetwork maybe we can just get it without searching for it? Might reduce more gas?

steven2308: I was too afraid to remove it in case we ever use more than one and we forget to change it :grimacing:

src/contracts/middleware/Middleware.sol
steven2308: Done

src/contracts/middleware/OBaseMiddlewareReader.sol
steven2308: Also, it's just called once, I don't think it's worth it

NindoK: xD I dunno if it's worth, that was my question xD can you check real quick? It's still a view function, so if the reduction is 10 gas who cares xD

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -735,4 +753,73 @@ contract OBaseMiddlewareReader is
...
+
+    /**
+     * @notice Optimized version of _getOperatorPowerAt that only gets the power for a single subnetwork
+     * @param timestamp The timestamp to check
+     * @param operator The operator address
+     * @param vaults The list of vault addresses
+     * @param subnetwork The subnetwork identifier
+     * @return power The total power amount at the timestamp
+     */
+    function _getOperatorPowerAt(
```
NindoK, line 812: Do we really need this? Can't we use the previous one with an array of subnetworks, but the subnetworks is only 1?

test/integration/Full.t.sol
```diff
@@ -2102,6 +2104,129 @@ contract FullTest is Test {
...
+            adminFeeSetRoleHolder: tanssi,
+            implementation: stakerRewardsImpl
+        });
+
+        for (uint256 i; i < maxVaults - activeSharedVaults - 1; i++) {
+            (address vault,,) = deployVault.createBaseVault(params);
+            middleware.registerSharedVault(address(vault), stakerRewardsParams);
+        }
+        vm.warp(block.timestamp + VAULT_EPOCH_DURATION + 1);
+        activeSharedVaults = middlewareReader.activeVaults().length;
```
NindoK, line 2140: I thiknk here we are missing an assert with MAX_LIMIT?

test/integration/Full.t.sol
```diff
@@ -2102,6 +2104,129 @@ contract FullTest is Test {
...
+        vm.expectRevert(IMiddleware.Middleware__TooManyActiveVaults.selector);
+        middleware.registerSharedVault(address(vaultAfterLimit), stakerRewardsParams);
+
+        vm.stopPrank();
+    }
+
+    function testWhenOperatorManagesToBeActiveInTooManyVaultsThenItIsIgnored() public {
+        vm.warp(block.timestamp + NETWORK_EPOCH_DURATION + 1);
+
+        // Before everything happens, the operator 1 is already registered in a vault and has power. Once he is in too many vaults his power must become 0 even if no stake is removed.
```
NindoK, line 2152: Let's specify it's 0, mainly because it goes out of gas

src/contracts/middleware/OBaseMiddlewareReader.sol
steven2308: My version is removing a for and caching the length of the vaults array. We don't really "need" any of this, but if we are optimizing why do it halfway?

test/integration/Full.t.sol
steven2308: If he is over the limit, he will be fully ignored. Otherwise it will revert later when slashing or distributing rewards.

steven2308: > Let's specify it's 0, mainly because it goes out of gas

The reader code already explains why it's ignored.

test/integration/Full.t.sol
steven2308: I really see no need. I'm creating vaults based on the max limit, and I'm testing it reverts once it goes over.

NindoK: Ye but it's just an assert and it makes sure that there's no issue in the flow when creating all these vaults. A check more that doesn't cost us antyhing

src/contracts/middleware/OBaseMiddlewareReader.sol
NindoK: Because it's another function to keep track of, while we can reuse what we already have by just passing an array with a single value. If we can reuse code might makes sense to do it. Also I dunno how much would it be the cost optimization here

NindoK: Run some tests. We keep your version it actually optimizes more than I thought for an array

test/integration/Full.t.sol
steven2308: Was getting active vaults instead of active shared vaults. Added the assertion too.

PR #142, opened 2025-07-15
[MD-1133] Define hints per vault on operatorRewards.claimRewards

steven2308:
From the data sent to operatorRewards.claimRewards we used to take hints to send to the staker contracts. However it would pretend to use the same hints on each staker rewards contract which may lead to out of bound errors on the subjacent Checkpoints data structure. So we changed the expected encoded data to include admin fee and an Array of VaultHints, which includes the vault and the hints to get active stake and active shares. 

A helper contract was added to calculate the data for free (in a view method) and then use on the claim call. Using hints provided by the contract provides significant savings, though in some cases it actually slightly increases the gas cost, I will continue researching this case.

Here are some examples of running tests without and with hints.

```
testDistributingAndClaimingRewardsForMultipleEpochs
  claimRewards gas: 545716->394183 27.76%
  claimRewards gas: 218292->220866 -1.17%
  claimRewards gas: 295316->297783 -0.83%

testRewardsAreDistributedCorrectlyForOperator1
  claimRewards gas: 847489->658936 22.24%

testRewardsAreDistributedCorrectlyForOperator2
  claimRewards gas: 545716->394183 27.76%
```

Some design decisions:
1. Vault is included in the new Struct to check against the current processing vault. If it doesn't match we revert instead of using empty hints, since we understand the user is trying to use hints but did not generate the correct data so he'd rather revert on simulation than pay extra gas.
2. We use an array and force order so we don't have to store a mapping in storage, even if temporarily. This would be costly. Transient storage is only available for value types, not for mappings.  The helper contract compensates any difficulty generating the right hints data.


src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -310,12 +310,15 @@ contract ODefaultOperatorRewards is
         bytes calldata data
     ) private {
         OperatorRewardsStorage storage $ = _getOperatorRewardsStorage();
+        // Originaly the 2nd and 3rd param are hints to optimize caching on the staker rewards contract. However it would pretend to use the same hints on each staker rewards contract which may lead to out of bound errors on the subjacent Checkpoints data structure. So we clean the hints and re-encode them.
+        (uint256 maxAdminFee,,) = abi.decode(data, (uint256, bytes, bytes));
+        bytes memory cleanData = abi.encode(maxAdminFee, new bytes(0), new bytes(0));
```
NindoK, line 315: The idea is to have hints passed by operators to speed lookups. In this case we need an array of data rather than a single bytes. So we shouldn't set `new bytes(0)` but rather take each hint for each vault

NindoK: Yes please

NindoK: But you concur that the error wouldn't happen if the hints are correct? So this is not actually solving if we need to redo it in another PR? Btw yes they can send a pair it makes sense

src/contracts/rewarder/ODefaultOperatorRewards.sol
```diff
@@ -223,7 +223,7 @@ contract ODefaultOperatorRewards is
             operatorVaults, totalVaults, epochStartTs, operator, middlewareAddress, stakerAmount
         );
 
-        _distributeRewardsPerVault(epoch, eraIndex, tokenAddress, totalVaults, operatorVaults, amountPerVault, data);
+        _distributeRewardsForEachVault(epoch, eraIndex, tokenAddress, totalVaults, operatorVaults, amountPerVault, data);
```
NindoK, line 226: Per has the same meaning xD

src/contracts/rewarder/RewardsPreparer.sol
```diff
@@ -0,0 +1,69 @@
...
+    IODefaultOperatorRewards public immutable i_operatorRewards;
+    IVaultHints public immutable i_vaultHints;
+
+    constructor(address middleware, address operatorRewards, address vaultHints) {
+        i_middlewareReader = IOBaseMiddlewareReader(middleware);
+        i_operatorRewards = IODefaultOperatorRewards(operatorRewards);
+        i_vaultHints = IVaultHints(vaultHints);
+    }
+
+    function getDataForOperatorClaimRewards(
```
NindoK, line 35: Love it. Good one

NindoK:
LFD. Looks awesome. Good job

src/contracts/rewarder/ODefaultOperatorRewards.sol
steven2308: I had to create method for a single vault due to stack too deep, so renamed this to make it more evident that one is for each and the other is for a single one :grimacing:

PR #145, opened 2025-07-17
[MD-1205] - Aggregator Oracle Proxy to support multiple oracles

NindoK:
This PR introduces a proxy contract, `AggregatorV3Interface` compatible, to integrate the DIA oracle into the project. This approach allows to use DIA price feed without requiring any changes to the existing codebase.

### DIFF
--- a/script/DeployRewards.s.sol
+++ b/script/DeployRewards.s.sol
@@ -83,8 +83,6 @@ contract DeployRewards is Script {
     ) public returns (address) {
         if (!isTest) {
             vm.startBroadcast(broadcaster());
-        } else {
-            vm.startPrank(owner_);
         }
         stakerRewardsFactory = new ODefaultStakerRewardsFactory(
             vaultFactory, networkMiddlewareService, operatorRewardsAddress, network, owner_
@@ -93,8 +91,6 @@ contract DeployRewards is Script {
 
         if (!isTest) {
             vm.stopBroadcast();
-        } else {
-            vm.stopPrank();
         }
 
         return address(stakerRewardsFactory);


src/contracts/oracle-proxy/AggregatorV3Proxy.sol
```diff
@@ -0,0 +1,61 @@
...
+    constructor(address _aggregator, string memory _pairSymbol) {
+        if (_aggregator == address(0) || bytes(_pairSymbol).length == 0) {
+            revert AggregatorV3Proxy__InvalidData();
+        }
+        aggregator = IDIAOracleV2(_aggregator);
+        pairSymbol = _pairSymbol;
+    }
+
+    function decimals() external pure override returns (uint8) {
+        return 8;
```
steven2308, line 25: Let's include this in the constructor so we can use it for others in the future without any change

src/contracts/oracle-proxy/AggregatorV3Proxy.sol
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.25;
+
+//**************************************************************************************************
+//                                      CHAINLINK
+//**************************************************************************************************
+import {AggregatorV3Interface} from "@chainlink/shared/interfaces/AggregatorV2V3Interface.sol";
+import {IDIAOracleV2} from "src/interfaces/oracles/IDIAOracleV2.sol";
+
+contract AggregatorV3Proxy is AggregatorV3Interface {
```
steven2308, line 10: Maybe call it something with DIA?  like `AggregatorV3DIAProxy`

test/integration/Middleware.t.sol
```diff
@@ -990,6 +999,74 @@ contract MiddlewareTest is Test {
...
+        _depositToVault(Vault(tanssiVaultAddress), operator5, OPERATOR_5_STAKE_TANSSI, tanssiCollateral);
+
+        vm.warp(NETWORK_EPOCH_DURATION + SLASHING_WINDOW - 1);
+        uint48 currentEpoch = middleware.getCurrentEpoch();
+        Middleware.ValidatorData[] memory validators = _validatorSet(currentEpoch);
+
+        // Total deposit is 300 USD, it should be normalized to 18 decimals
+        uint256 totalPowerByShares = 150 ether;
+        // Only 2 operators participate in the USD vaults, so each has half of the power.
+        uint256 totalPowerOperator = totalPowerByShares / 2;
```
steven2308, line 1064: I'm surprised you don't need to multiply by oracle price and divide by decimals, how is it possible?
Also, the to comments above seem off

steven2308 (changes requested):
Really minor things. Only doubt is on the power check how it even works :sweat_smile:

test/integration/Middleware.t.sol
NindoK: Probably because I set the price to 1 👀. Probably better some strange number

steven2308: Oh yeah, let's definitely test with a different price and decimals.

steven2308 (approved):
LFD!

PR #169, opened 2025-08-01
Improvements on check and perform upkeep

steven2308:
1. Only top 59 operators by power are sent to substrate, to comply with max perform size of 2k bytes.
2. Uses all operators instead of only active to reduce drastically the cost of check and perform up keep. The cost of performing upkeep is now stable, regardless of the number of operators.

Gas report [here](https://docs.google.com/spreadsheets/d/1_MbAaiaEYeMlhIXGHi8mZAuZHclZVktEkvrhAu6Zq2Q/edit?gid=0#gid=0)
Estimations with gas price at 2 GWEI and Eth price at 3.5k USD.
image


tevrat-aksoy:
@NindoK 

Can you check this example scenario? Is this a possible situation or am I missing something?

Total 5 operators: A, B, C, D, E.

Operator C has paused (inactive) from the previous epoch  but is still present in the _operators set, not unregistered

First Call to auxiliaryCheckUpkeep :  auxiliaryCheckUpkeep returns 4 operators ( C is not included) 

Call to performUpkeep: performUpkeep updates index to 4

Next call to auxiliaryCheckUpkeep:  pending would start from 1 (5-4) which should have procesed all operators should have been 0

`        uint256 pendingOperatorsToCache = operatorsLength_ - cacheIndex;
`

So it would try to process 4. element which is E and already processed. so performUpkeep would fail .

New operators added : A, B, C, D, E, F, G, H

C is unregistered: Operator indexes updated -> A, B, H, D, E, F, G

auxiliaryCheckUpkeep would start from index 4 again and proceses E, F , G

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -725,58 +725,75 @@ contract OBaseMiddlewareReader is
...
         // Check if cache is still not filled with the current epoch validators
         if (pendingOperatorsToCache > 0) {
             uint256 maxNumOperatorsToCheck = Math.min(pendingOperatorsToCache, MAX_OPERATORS_TO_PROCESS);
-            IMiddleware.ValidatorData[] memory validatorsData =
-                _getValidatorDataForOperators(maxNumOperatorsToCheck, cacheIndex, currentEpochStartTs, activeOperators_);
+            (IMiddleware.ValidatorData[] memory validatorsData, bool atLeastOneActive) =
+                _getValidatorDataForOperators(maxNumOperatorsToCheck, cacheIndex, currentEpochStartTs, operators);
+
+            // This is the case were 100% of the operators are inactive, so we don't need to send anything
+            if (operatorsLength_ <= MAX_OPERATORS_TO_SEND && !atLeastOneActive) {
```
NindoK, line 747: I think this should be an `||`?

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -725,58 +725,75 @@ contract OBaseMiddlewareReader is
...
+            (address operator,,) = operators.at(i);
             bytes32 operatorKey = abi.decode(operatorKey(operator), (bytes32));
-            uint256 operatorPower = _optmizedGetOperatorPowerAt(timestamp, sharedVaults, subnetwork, operator);
-            validatorsData[i - cacheIndex] = IMiddleware.ValidatorData({key: operatorKey, power: operatorPower});
+            if (operators.wasActiveAt(timestamp, operator)) {
+                atLeastOneActive = true;
+                uint256 operatorPower = _optmizedGetOperatorPowerAt(timestamp, sharedVaults, subnetwork, operator);
+                validatorsData[i - cacheIndex] = IMiddleware.ValidatorData({key: operatorKey, power: operatorPower});
+            } else {
+                validatorsData[i - cacheIndex] = IMiddleware.ValidatorData({key: operatorKey, power: 0});
```
NindoK, line 795: Why do we add it? If it wasn't active, just don't add it

NindoK: Nothing just checked that if we don't do it we are going to have mismatches between cache index and length

NindoK:
> @NindoK
> 
> Can you check this example scenario? Is this a possible situation or am I missing something?
> 
> Total 5 operators: A, B, C, D, E.
> 
> Operator C has paused (inactive) from the previous epoch but is still present in the _operators set, not unregistered
> 
> First Call to auxiliaryCheckUpkeep : auxiliaryCheckUpkeep returns 4 operators ( C is not included)
> 
> Call to performUpkeep: performUpkeep updates index to 4
> 
> Next call to auxiliaryCheckUpkeep: pending would start from 1 (5-4) which should have procesed all operators should have been 0
> 
> ` uint256 pendingOperatorsToCache = operatorsLength_ - cacheIndex;`
> 
> So it would try to process 4. element which is E and already processed. so performUpkeep would fail .
> 
First call in this case would return also C, but it will not be active. We will add it in the validatorsData array but with 0 power.

> New operators added : A, B, C, D, E, F, G, H
> 
> C is unregistered: Operator indexes updated -> A, B, H, D, E, F, G
> 
> auxiliaryCheckUpkeep would start from index 4 again and proceses E, F , G

If C is unregistered index will be 0 since it will be a new epoch. (This is mandatory. *You can't be unregistered/pause in the current epoch*)

src/contracts/middleware/OBaseMiddlewareReader.sol
steven2308: No. If operatorsLength_ > MAX_OPERATORS_TO_SEND, it could be that the first batch is all inactive and I still need to return them so it's called again or the next batch.
I can only quit early if I know for sure all of they are all inactive and they are all in this batch.

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -782,17 +784,16 @@ contract OBaseMiddlewareReader is
...
-        uint256 operatorsLength_ = operators.length();
 
         for (uint256 i = cacheIndex; i < cacheIndex + maxNumOperatorsToCheck && i < operatorsLength_;) {
-            (address operator,,) = operators.at(i);
-            bytes32 operatorKey = abi.decode(operatorKey(operator), (bytes32));
-            if (operators.wasActiveAt(timestamp, operator)) {
+            (address operator, uint48 enabled, uint48 disabled) = operators.at(i);
+            validatorsData[i - cacheIndex].key = abi.decode(operatorKey(operator), (bytes32));
+
+            if (enabled < timestamp && (disabled == 0 || disabled >= timestamp)) {
```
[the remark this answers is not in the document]
steven2308: I would extract it only if we ever needed somewhere else. Otherwise it's just an extra JUMP call, so a bit more gas with no good reason to.

test/integration/Middleware.t.sol
```diff
@@ -1646,16 +1711,14 @@ contract MiddlewareTest is Test {
         address offlineKeepers = makeAddr("offlineKeepers");
 
         vm.startPrank(offlineKeepers);
-        (bool upkeepNeeded, bytes memory performData) = middleware.checkUpkeep(hex"");
```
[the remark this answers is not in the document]
steven2308: Glad you catched it, I actually removed it because it was failing but wanted to check later. Will do after testing the zero power removal. Let's not resolve it for now.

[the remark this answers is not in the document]
steven2308: No, but I just found the reason. In the two tests I removed this check from, we are testing with 100% operators.
Previously in the init call they were all inactive and since we checked them all we would know no upkeep was needed.
Now since we only check the first batch of 20, even if we find the full batch is inactive, we cannot conclude that all the other batches are, so we must still return that the `performUpkeep` is needed.

This might seem inefficient, but keep in mind that we will rarely have inactive operators, this is just a corner case on the way we test, and specifically because we are doing this check as soon as they are added. Even for these tests, on the next epoch they will all be active.

So I would keep those lines removed.

PR #170, opened 2025-08-04
[H-03] Adds test showing operators cannot change during an epoch

NindoK:
Operators can't change during an epoch, hence checkUpkeep always return the active at the start of epoch

https://github.com/PashovAuditGroup/Tanssi_July25_MERGED/issues/12

### DIFF
--- a/test/integration/Middleware.t.sol
+++ b/test/integration/Middleware.t.sol
@@ -560,7 +560,7 @@ contract MiddlewareTest is Test {
         Middleware.ValidatorData[] memory validatorsPreviousEpoch =
             OBaseMiddlewareReader(address(middleware)).getValidatorSet(previousEpoch);
 
-        vm.warp(block.timestamp + NETWORK_EPOCH_DURATION + 2);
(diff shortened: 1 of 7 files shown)


test/integration/Middleware.t.sol
```diff
@@ -1657,6 +1658,63 @@ contract MiddlewareTest is Test {
...
+            _calculateOperatorPower(totalPowerVaultSlashable, totalFullRestakePower, 0);
+        (uint256 totalOperator3PowerAfter,) =
+            _calculateOperatorPower(totalPowerVault + totalPowerVaultSlashable, totalFullRestakePower, 0);
+
+        assertEq(operator1Power, totalOperatorPowerAfter);
+        assertEq(operator2Power, totalOperator2PowerAfter);
+        assertEq(operator3Power, totalOperator3PowerAfter);
+
+        vm.startPrank(owner);
+        middleware.pauseOperator(operator);
```
steven2308, line 1693: Let's try unregistering the operator immediately with an expertRevert.

NindoK: We already have this in many other tests. I don't think it's needed here too

steven2308 (approved):

PR #174, opened 2025-08-04
[L-04] Collision risk from immutable to storage variable conversion

steven2308:
Fixes https://github.com/PashovAuditGroup/Tanssi_July25_MERGED/issues/4
Moves the two variables to the storage. Updates code and tests accordingly.

Since we already called reinitializeRewards on Dancelight, we had to change the reinitializer version to 3 to be able to call it again. This won't affect mainnet, we can go from 1 to 3.

### DIFF
--- a/src/contracts/middleware/MiddlewareStorage.sol
+++ b/src/contracts/middleware/MiddlewareStorage.sol
@@ -58,7 +58,7 @@ abstract contract MiddlewareStorage {
      * @notice Get the operator rewards contract address
      * @return operator rewards contract address
      */
-    function getOperatorRewards() public view returns (address) {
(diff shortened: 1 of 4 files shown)


src/contracts/middleware/MiddlewareStorage.sol
```diff
@@ -56,13 +58,19 @@ abstract contract MiddlewareStorage {
      * @notice Get the operator rewards contract address
      * @return operator rewards contract address
      */
-    address public i_operatorRewards;
+    function getOperatorRewards() public view returns (address) {
```
NindoK, line 61: `getOperatorRewardsAddress` it can be misleading otherwise. Or just call it i_operatorRewards() as if it was a simple variable

PR #187, opened 2025-08-12
Added forwarder addresses

NindoK:


NindoK:
> Shall this be on the confluence docs too?

Added, even though it's not needed

PR #193, opened 2025-08-25
Make mainnet fork tests more dynamic

steven2308:
1. Removes all operations on setup different to loading contracts and caching data.
2. Where possible, makes code independent from known vaults and operators.
3. Removes inactive operators.
4. Updates operator keys for Pier 2 and Alchemy.
5. Regenerates 3 proofs to test reward distribution for operators and stakers.
6. Tests slashing and executing (if needed) for every active vault.

Logic for operators is tested fully, since they are simply addresses from Middleware perspective, there is nothing pending to test for them.

Even though multiple tests used fixed vaults or operators this is left in place in two cases:
1. To test rewards claiming, since generating a proof on the fly is not viable from a smart contract.
2. To test withdrawals, since some vaults are private and there is now way to find the despositor whitelister address from a smart contract.


contract-addresses/tanssi.json
```diff
@@ -213,50 +213,20 @@
...
       "name": "Pier Two",
-      "substrateAddress": "5HRErQcsVBvG1SArk7Xp1Yd8VXSAVGvNHqFnZtGRybCAYotP",
+      "substrateAddress": "5GN5st9G5PzoWjwC3FPWkSVkcfvqGxvhCXuH4fhSV6AsNwwR",
       "evmAddress": "0x51B6D824bd35AeD4FD1a9E253E41Dc7C9feeFa30",
-      "operatorKey": "0xecd66a5fac2d2b6abbc31562058b30dc8eaffaf229b19c0bacd170995dfd0e5e"
-    },
-    {
-      "name": "P2P",
-      "substrateAddress": "5GYtpbcHHjArLe5PfG2R412JnufdHibPQg2jGCLoEJQC4RVU",
-      "evmAddress": "0x087c25f83ED20bda587CFA035ED0c96338D4660f",
```
NindoK, line 227: I'm not entirely sure we should kick them out from the tanssi.json. They are registered, but just paused. Taking them out makes it more like saying: you are not in any way present in the middleware. I'd keep it as of now, since we will unpause them in the short term and making a PR for this feels redundant

script/HelperConfig.s.sol
```diff
@@ -98,35 +98,24 @@ contract HelperConfig is Script {
...
-        OperatorData operator8CP0XStakrspace;
-        OperatorData operator9HashkeyCloud;
-        OperatorData operator10Alchemy;
-    }
-
-    struct OperatorConfigB {
-        OperatorData operator11Opslayer;
-        OperatorData operator12TanssiFoundation;
+        OperatorData operator2Nodeinfra;
+        OperatorData operator3CP0XStakrspace;
```
NindoK, line 106: Revert this to previous solution

test/unit/DeployProduction.t.sol
```diff
@@ -111,13 +110,13 @@ contract DeployProductionTest is Test {
         assertNotEq(vaultsConfigB.gauntletRestakedWstETH.delegator, address(0));
         assertNotEq(vaultsConfigB.gauntletRestakedWstETH.slasher, address(0));
 
-        assertNotEq(operatorsA.operator1PierTwo.name, "");
-        assertNotEq(operatorsA.operator1PierTwo.evmAddress, address(0));
-        assertNotEq(operatorsA.operator1PierTwo.operatorKey, bytes32(0));
+        assertNotEq(operators.operator1PierTwo.name, "");
+        assertNotEq(operators.operator1PierTwo.evmAddress, address(0));
```
NindoK, line 114: Revert

contract-addresses/tanssi.json
steven2308: They either never produced blocks on a testnet or did not complete the process.
I'm ok with keeping them in tanssi.json, but it makes no sense to have them in tests. We want to test with real life data, and for these guys I had to mock a ton of things.
WDYT?

test/fork/mainnet/Full.t.sol
```diff
@@ -1388,7 +443,7 @@ contract FullTest is Test {
             abi.decode(performData, (uint8, uint48, IMiddleware.ValidatorData[]));
         assertEq(epoch, currentEpoch);
         assertEq(command, middleware.CACHE_DATA_COMMAND());
-        assertEq(validatorsData.length, TOTAL_OPERATORS);
+        assertGe(validatorsData.length, TOTAL_OPERATORS);
```
NindoK, line 446: How can it be greater?

contract-addresses/tanssi.json
NindoK: We will use them. They are getting onboarded as of now! Keeping the struct will help us in not changing the code later on

steven2308: Why will we use them? 
For us operators are nothing but an EVM address. There is nothing to test there.

PR #197, opened 2025-09-02
Sends operator rewards only once when doing batch claim.

steven2308:
Produces 0.50% in a batch of 3. Bigger batches will produce bigger savings.

This tests are with an operator in only 2 vaults and doing a batch claim for 3 eras:
**Before:**
  claimRewards gas Hints: `1513792`
  claimRewards gas No Hint: `1691283` 

**Noew:**
  claimRewards gas Hints: `1506164`
  claimRewards gas No Hint: `1683657`


src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -146,6 +146,11 @@ contract ODefaultStakerRewards is
         address tokenAddress
     ) external view override returns (uint256 amount) {
         StakerRewardsStorage storage $ = _getStakerRewardsStorage();
+
+        if ($.activeSharesCache[epoch] == 0) {
```
NindoK, line 150: This might not be true. claimable could be > 0, but we still didn't update cache hence this will be always 0 until claimed by someone

NindoK (approved):

PR #202, opened 2025-09-05
Implements batchClaimRewardsAndRestake on StakerRewards.

steven2308:
You can only restake if the collateral token is the same as the rewards token.

### DIFF
--- a/src/contracts/rewarder/ODefaultStakerRewards.sol
+++ b/src/contracts/rewarder/ODefaultStakerRewards.sol
@@ -294,6 +294,43 @@ contract ODefaultStakerRewards is
         address tokenAddress,
         bytes[] calldata activeSharesOfHints
     ) external {
+        batchClaimRewardsAndRestake(recipient, epochs, tokenAddress, activeSharesOfHints, 0);
+    }
+
+    /**
+     * @inheritdoc IODefaultStakerRewards
+     */
+    function batchClaimRewardsAndRestake(
+        address recipient,
+        uint48[] calldata epochs,
+        address tokenAddress,
+        bytes[] calldata activeSharesOfHints,
+        uint48 restakePercentageBps
+    ) public {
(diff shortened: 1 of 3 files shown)


src/contracts/rewarder/ODefaultStakerRewards.sol
```diff
@@ -294,6 +294,43 @@ contract ODefaultStakerRewards is
...
+        bytes[] calldata activeSharesOfHints,
+        uint48 restakePercentageBps
+    ) public {
+        uint256 totalAmount = _batchClaimRewards(recipient, epochs, tokenAddress, activeSharesOfHints);
+
+        uint256 restakeAmount = totalAmount.mulDiv(restakePercentageBps, 10_000);
+        uint256 claimAmount = totalAmount - restakeAmount;
+
+        if (claimAmount != 0) {
+            IERC20(tokenAddress).safeTransfer(recipient, claimAmount);
```
NindoK, line 316: I love this. You can just restake only a percentage. So much modularity in terms of UX. LOVE IT

test/integration/Full.t.sol
```diff
@@ -2215,6 +2212,92 @@ contract FullTest is Test {
...
+    function testStakerRewardsAreDistributedCorrectlyWhenBatchClaimedAndRestaked50Percent() public {
+        (
+            uint256 previousFreeBalance,
+            uint256 previousStakedBalance,
+            uint256 stakerRewardsOperator1Vault1,
+            uint256 newFreeBalance,
+            uint256 newStakedBalance
+        ) = _prepareRestakeTest(5000);
+
+        assertApproxEqAbs(newFreeBalance, previousFreeBalance + stakerRewardsOperator1Vault1 / 2, 1);
```
NindoK, line 2262: With mulDiv you can't get the precision?

steven2308: The rewards were odd. So one half has to take 1 more wei inevitably. Even with `mulDiv`

PR #206, opened 2025-09-22
[MD-1360] - Added caching steps for vault powers in order to understand how much should go to each vault staker rewards

NindoK:
This PR introduces a new way to handle distribution of rewards. It doesn't implement it directly, but gives the foundation to support it.

CL Automation now interact with the middleware like it's a state machine where there are 2 main states for caching:

- `ExpectingOperatorData`
- `ExpectingVaultData`

The previous operator powers caching stays the same.

While this PR introduces the vault power caching, by going through all the operators and getting their powers for each their vault they are in. In this way it's possible to calculate the total power per vault.

This will allow to calculate in an upcoming PR the rewards distribution per stakers by calculating:

`TOTAL_AMOUNT.mulDiv(VAULT_POWER / TOTAL_VAULTS_POWER)`

This should allow to have one single staker reward contract!

### DIFF
--- a/src/contracts/middleware/OBaseMiddlewareReader.sol
+++ b/src/contracts/middleware/OBaseMiddlewareReader.sol
@@ -746,11 +746,27 @@ contract OBaseMiddlewareReader is
         }
 
         uint256 cacheIndex = cache.epochToCacheIndex[epoch];
-        uint256 pendingOperatorsToCache = operatorsLength_ - cacheIndex;
 
+        IMiddleware.CachingState cacheState = cache.epochToNextExpectedCacheCommand[epoch];
+
+        uint256 pendingOperatorsToCache = operatorsLength_ - cacheIndex;
         // Check if cache is still not filled with the current epoch validators
         if (pendingOperatorsToCache > 0) {
             uint256 maxNumOperatorsToCheck = Math.min(pendingOperatorsToCache, MAX_OPERATORS_TO_PROCESS);
+            if (cacheState == IMiddleware.CachingState.ExpectingVaultData) {
+                (IMiddleware.VaultPower[] memory vaultsPower, bool atLeastOne, uint256 operatorsProcessed) =
+                _getOperatorPowerPerVault(
+                    maxNumOperatorsToCheck, cacheIndex, currentEpochStartTs, operators, operatorsLength_
+                );
+
+                // This should be useless since it is always checked on the validator operators process. Impossible to reach this state
+                if (operatorsLength_ <= MAX_OPERATORS_TO_SEND && !atLeastOne) {
+                    return (false, hex"");
+                }
+
+                return (true, abi.encode(CACHE_VAULTS_DATA_COMMAND, epoch, operatorsProcessed, vaultsPower));
+            }
+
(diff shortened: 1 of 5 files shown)


src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -746,11 +746,27 @@ contract OBaseMiddlewareReader is
...
         // Check if cache is still not filled with the current epoch validators
         if (pendingOperatorsToCache > 0) {
             uint256 maxNumOperatorsToCheck = Math.min(pendingOperatorsToCache, MAX_OPERATORS_TO_PROCESS);
+            if (cacheState == IMiddleware.CachingState.ExpectingVaultData) {
+                (IMiddleware.VaultPower[] memory vaultsPower, bool atLeastOne, uint256 operatorsProcessed) =
+                _getOperatorPowerPerVault(
+                    maxNumOperatorsToCheck, cacheIndex, currentEpochStartTs, operators, operatorsLength_
+                );
+
+                // This should be useless since it is always checked on the validator operators process. Impossible to reach this state
```
steven2308, line 762: I think it's reachable.
1. If `!atLeastOne` that means all operators in this batch had no power. 
2. If `operatorsLength_ <= MAX_OPERATORS_TO_SEND` that means they all fit in one batch.
3. So we have all operators here, and all have no power. We can do an early exit since there's no point on sending anything to substrate. But that opens a question: 

What should we tell substrate if no operators have power? 😨

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -780,6 +796,121 @@ contract OBaseMiddlewareReader is
...
+        uint48 timestamp,
+        PauseableEnumerableSet.AddressSet storage operators,
+        uint256 operatorsLength_
+    )
+        private
+        view
+        returns (IMiddleware.VaultPower[] memory vaultsPower, bool atLeastOneActive, uint256 operatorsProcessed)
+    {
+        VaultManagerStorage storage $ = _getVaultManagerStorage();
+        address[] memory sharedVaults = $._sharedVaults.getActive(timestamp);
```
steven2308, line 811: What about operator specific vaults?

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -780,6 +796,121 @@ contract OBaseMiddlewareReader is
...
+    }
+
+    function _extractVaultsPower(
+        IMiddleware.VaultPower[] memory vaultsPower,
+        uint256 totalVaultsIndex_,
+        address vault,
+        uint256 power
+    ) private pure returns (uint256 totalVaultsIndex) {
+        // Inner accumulation loop
+        for (uint256 k; k < vaultsPower.length; ++k) {
```
steven2308, line 874: I'm guessing this was extracted and `k` used to be needed, but can we use `i` instead of `k` now?

steven2308 (commented):
Sorry for the delay, I needed a very clear mind to review it and didn't have it until now 😅 
I think we are missing operator specific vaults, other than that it looks good. However my concern about being able to get the same rewards for each vault remains, so I'd leave this unmerged until proven.

PR #225, opened 2025-12-18
Fixes foundry to v1.3.0 and runs new format.

steven2308:


script/Checks.s.sol
NindoK, line 1: What's up with this?

script/DeployVault.s.sol
```diff
@@ -66,21 +66,15 @@ contract DeployVault is Script {
...
 
     function createSlashableVault(
         CreateVaultBaseParams memory params
     ) public returns (address, address, address) {
         return _createVault({
-            params: params,
-            withSlasher: true,
-            slasherIndex: VaultManager.SlasherType.INSTANT,
-            vetoDuration: 0
+            params: params, withSlasher: true, slasherIndex: VaultManager.SlasherType.INSTANT, vetoDuration: 0
```
NindoK: Wut why? It pushes everything in multiple lines and this in a single line? 😂

src/contracts/middleware/OBaseMiddlewareReader.sol
```diff
@@ -55,6 +55,7 @@ import {MiddlewareCRELogic} from "src/contracts/libraries/MiddlewareCRELogic.sol
  * management capabilities that can be extended with additional functionality.
  */
 contract OBaseMiddlewareReader is
+
```
NindoK: Wut

NindoK (approved):

script/Checks.s.sol
steven2308: That was not meant to be pushed :laughing: 
I used it for running stuff from local. Taking it back

script/DeployVault.s.sol
steven2308: Yeah.. non sense really.

NindoK: Can we fix it? I think we can add 'multiple' to the toml

steven2308: Can't find a way, only related changeable thing is line length. 
Or we switch formatter but meh..

NindoK (approved):

